PR #240, opened 2024-11-09
AV-90385 Azure cluster - ultra and premium chagNes

PaulomeeCb:
## Jira

* AV-90385

## Description

1. After this change - https://github.com/couchbasecloud/couchbase-cloud/pull/35815, the condition was removed to throw error when user adds iops and storage for Azure premium-type disks. So now in case the user adds the iops and storage for Premium, they should be ignored, all other types of clusters remain same.
2. Also, the deployment and update of Ultra-type azure clusters were not working from Terraform before. 
3. The update of Ultra -> Premium or Premium -> Ultra was failing.

This change ensures all the above are resolved, the Azure cluster is deployed and can be updated properly for Azure clusters.

TODO:
- Add testing proofs
- Add examples to .tf and Readme for azure cluster

Please include a summary of the fix/feature/change, including any relevant motivation and context.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue). Please, add the "bug" label to the PR.
- [ ] Documentation fix/enhancement

- [x] Manually tested
- [ ] Unit tested
- [ ] Acceptance tested

### Testing

Will add ss/ recording.

<details open>
  <summary>Testing</summary>
  
</details>

## Required Checklist:

- [ ] I have added any necessary documentation (if required)
- [ ] I have run make fmt and formatted my code
- [ ] I have made sure that no schema field is marked with both requiresReplace and computed
## Further comments

### DIFF
--- a/examples/cluster/azure/variables.tf
+++ b/examples/cluster/azure/variables.tf
@@ -45,7 +45,7 @@ variable "disk" {
   description = "All nodes' disk configuration"
 
   type = object({
-    size          = optional(number)
+    storage          = optional(number)
     type          = string
     iops          = optional(number)
     autoexpansion = optional(bool)
--- a/internal/resources/cluster.go
+++ b/internal/resources/cluster.go
@@ -210,6 +210,16 @@ func (c *Cluster) Create(ctx context.Context, req resource.CreateRequest, resp *
 		if clusterapi.AreEqual(plan.ServiceGroups[i].Services, serviceGroup.Services) {
 			refreshedState.ServiceGroups[i].Services = plan.ServiceGroups[i].Services
 		}
+
+		// This is for the case when for Azure clusters - Premium disk types, if the user provides the storage and iops values (these values are ignored), then put the value provided in state file,
+		// as otherwise there will be a type mismatch with the actual value - i.e. the default values for premium disks.
+		if refreshedState.CloudProvider.Type == types.StringValue("Azure") {
+			if refreshedState.ServiceGroups[i].Node.Disk.Type != types.StringValue("Ultra") {
+				refreshedState.ServiceGroups[i].Node.Disk.Storage = plan.ServiceGroups[i].Node.Disk.Storage
+				refreshedState.ServiceGroups[i].Node.Disk.IOPS = plan.ServiceGroups[i].Node.Disk.IOPS
+			}
+		}
+
 	}
 
 	// Set state to fully populated data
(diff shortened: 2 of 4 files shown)


examples/cluster/azure/terraform.template.tfvars
```diff
@@ -29,3 +29,35 @@ support = {
   plan     = "basic"
   timezone = "PT"
 }
+
+
+# Example for Ultra disk type Azure Cluster:
```
Talina06, line 34: Can we make this a different example file? so that we don't need to comment it and the file will be usable as-is.

examples/cluster/azure/variables.tf
```diff
@@ -45,7 +45,7 @@ variable "disk" {
   description = "All nodes' disk configuration"
 
   type = object({
-    size          = optional(number)
+    storage       = optional(number)
```
Talina06, line 48: with the renaming of this, how do we ensure backwards compatibility?

IMO it is ok, but we need to be very clear that this gets mentioned properly in our release notes and we do a major version update...
FYI @rajagp @a-atri

internal/resources/cluster.go
```diff
@@ -423,6 +433,16 @@ func (c *Cluster) Update(ctx context.Context, req resource.UpdateRequest, resp *
 		if clusterapi.AreEqual(plan.ServiceGroups[i].Services, serviceGroup.Services) {
 			currentState.ServiceGroups[i].Services = plan.ServiceGroups[i].Services
 		}
+
+		// for Azure clusters - Premium disk types, if the user provides the storage and iops values (these values are ignored), then put the value provided in state file,
+		// as otherwise there will be a type mismatch with the actual value - i.e. the default values for premium disks.
+		if currentState.CloudProvider.Type == types.StringValue("Azure") {
```
Talina06, line 439: Azure should be a const, as I assume it will get used at multiple places.
same for other strings below.

internal/resources/cluster.go
```diff
@@ -673,6 +696,15 @@ func (c *Cluster) morphToApiServiceGroups(plan providerschema.Cluster) ([]cluste
...
 			}
+
+			//fmt.Printf("************************PAULOMEE Create/Update DISKAzure  %#v \n", diskAzure)
+			//b, _ := json.Marshal(diskAzure)
+			//fmt.Print(string(b))
+			//
+			//fmt.Printf("************************PAULOMEE Create/Update node  %#v \n", node)
+			//b, _ = json.Marshal(node)
+			//fmt.Print(string(b))
+
```
Talina06, line 707: rm

examples/cluster/azure/variables.tf
PaulomeeCb: I will rename to `size` for backwards compatibility

SaicharanCB (approved):
+1
tested here http://172.23.121.80/job/terraform-test/789/

SaicharanCB (approved):

internal/resources/cluster.go
```diff
@@ -835,3 +863,54 @@ func initializePendingClusterWithPlanAndId(plan providerschema.Cluster, id strin
 	}
 	return plan
 }
+
+func (c *Cluster) checkDisk(plan providerschema.Cluster) error {
```
l0n3star, line 901: This logic should be on V4 API not terraform.

l0n3star (changes requested):

internal/resources/cluster.go
PaulomeeCb: We had a long discussion on this and decided to not touch v4 API, and do the change on terraform. CC: @a-atri @rajagp @Talina06

l0n3star (approved):

internal/resources/cluster.go
l0n3star: On second thought API server already does disk validation.  So blocking from client is valid approach.

PR #294, opened 2025-04-30
AV-100893-Adding Generic function to add markdown description to the fields

PaulomeeCb:
## Jira

* AV-100893

## Description

Please include a summary of the fix/feature/change, including any relevant motivation and context.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue). Please, add the "bug" label to the PR.
- [X] Documentation fix/enhancement

- [X] Manually tested
- [ ] Unit tested
- [ ] Acceptance tested

### Testing

<details open>
  <summary>Testing</summary>
  
</details>

## Required Checklist:

## Further comments


internal/resources/attributes.go
```diff
@@ -60,20 +118,26 @@ func stringAttribute(fields []string, validators ...validator.String) *schema.St
...
+	if len(t) > 0 {
+		t[0](attribute)
+	}
+	return attribute
 }
 
 // stringDefaultAttribute sets the default values for a string field and returns the string attribute.
-func stringDefaultAttribute(defaultValue string, fields ...string) *schema.StringAttribute {
-	attribute := stringAttribute(fields)
+func stringDefaultAttribute(defaultValue string, fields []string, t ...Test[*schema.StringAttribute]) *schema.StringAttribute {
```
SaicharanCB, line 130: why are we using varidiac param here if only one value is being passed?

internal/resources/attributes.go
```diff
@@ -28,13 +29,70 @@ const (
...
+			v.MarkdownDescription = description
+		case *schema.NumberAttribute:
+			v.MarkdownDescription = description
+		case *schema.ListAttribute:
+			v.MarkdownDescription = description
+		}
+	}
+}
+
+func stringAttributeWithValidators(fields []string, t Test[*schema.StringAttribute], validators ...validator.String) *schema.StringAttribute {
```
SaicharanCB, line 58: It's better to combine this and string attribute function into one.

internal/resources/attributes.go
```diff
@@ -28,13 +29,75 @@ const (
...
+			v.MarkdownDescription = description
+		case *schema.SetAttribute:
+			v.MarkdownDescription = description
+		case *schema.Float64Attribute:
+			v.MarkdownDescription = description
+		case *schema.NumberAttribute:
+			v.MarkdownDescription = description
+		case *schema.ListAttribute:
+			v.MarkdownDescription = description
+		}
```
l0n3star, line 59: you shouldn't need to do type switch with generics

internal/resources/attributes.go
```diff
@@ -28,13 +29,75 @@ const (
...
+// It accepts an attribute of type schema.StringAttribute, schema.Int64Attribute,
+// schema.BoolAttribute, schema.SetAttribute, schema.Float64Attribute,
+// schema.NumberAttribute, or schema.ListAttribute, and a description string.
+// The function returns an AttributeFunc that modifies the attribute.
+func withMarkdown[T interface {
+	*schema.StringAttribute | *schema.Int64Attribute | *schema.BoolAttribute | *schema.SetAttribute |
+	*schema.Float64Attribute | *schema.NumberAttribute | *schema.ListAttribute
+}](description string) AttributeFunc[T] {
+	return func(s T) {
+		switch v := any(s).(type) {
```
l0n3star, line 44: no point in using generics if you have to do type switch

why not use interface ?

internal/resources/attributes.go
```diff
@@ -28,13 +29,75 @@ const (
 	deprecationMessage = "Remove this attribute's configuration as it no longer in use and the attribute will be removed in the next major version of the provider."
 )
 
+type AttributeFunc[T any] func(T)
```
l0n3star, line 32: should this be common to both datasources and resources ?

l0n3star:
please document validations you did

internal/resources/attributes.go
l0n3star: looking at this further, with go generics we can't access struct fields.  see https://github.com/golang/go/issues/48522.

so interface is better fit.

PR #295, opened 2025-04-30
AV-100893 - Adding markdown desc to fields - generics approach

PaulomeeCb:
## Jira

* [AV-100893](https://jira.issues.couchbase.com/browse/AV-100893)

## Description

Adding a Docs function to be used to add description for the fields.
Note: the resource files updated here are just for testing purposes, we need to add the proper description to all the fields as followed in the tickets under - https://jira.issues.couchbase.com/browse/AV-100781

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue). Please, add the "bug" label to the PR.
- [x] Documentation fix/enhancement

- [x] Manually tested
- [ ] Unit tested
- [ ] Acceptance tested

### Testing

<details open>
  <summary>Testing</summary>
How its appearing for clusters -

Screenshot 2025-05-01 at 10 38 35 AM

Screenshot 2025-04-30 at 11 04 04 AM

Backup fields where description is added-

Screenshot 2025-05-01 at 10 40 00 AM

Screenshot 2025-05-01 at 10 40 06 AM

How it would appear for fields in free-tier app service -

Screenshot 2025-05-01 at 10 41 42 AM

</details>

## Required Checklist:

## Further comments


internal/resources/attributes.go
```diff
@@ -28,6 +28,34 @@ const (
...
 
+// WithDescription sets the MarkdownDescription for the provided attribute.
+// It accepts an attribute of type schema.StringAttribute, schema.Int64Attribute,
+// schema.BoolAttribute, schema.SetAttribute, schema.Float64Attribute,
+// schema.NumberAttribute, or schema.ListAttribute, and a description string.
+// The function returns the modified attribute.
+func WithDescription[T interface {
+	*schema.StringAttribute | *schema.Int64Attribute | *schema.BoolAttribute | *schema.SetAttribute |
+		*schema.Float64Attribute | *schema.NumberAttribute | *schema.ListAttribute
+}](attr T, description string) T {
```
itsHabib, line 39: u should be able to wrap these in a type outside this call so that u can just do
```go
func WithDescription[T <CustomType](...) {...}
```

PaulomeeCb: Thanks for the suggestion! Done- https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/295/commits/ed227edb08974067c76cefa1e0e8904edbb39dbf

internal/resources/attributes.go
```diff
@@ -28,6 +28,34 @@ const (
...
+	switch v := any(attr).(type) {
```
l0n3star, line 43: since we're doing type switch, don't think generics is a good fit here.  

I just found with go generics we can't access struct fields. see https://github.com/golang/go/issues/48522.

seems like we should use interface instead.

internal/resources/attributes.go
```diff
@@ -28,6 +28,37 @@ const (
 	deprecationMessage = "Remove this attribute's configuration as it no longer in use and the attribute will be removed in the next major version of the provider."
 )
 
+// CustomType is a type alias that encapsulates the allowed attribute types.
+type CustomType interface {
```
itsHabib, line 32: nit: a name like `SchemaAttribute` would probably read better. in my previous comment i didnt explicitly mean to copy the name sorry :)

PaulomeeCb: Yes that makes sense

internal/resources/attributes.go
l0n3star: never mind just saw Michael's comment, so that is valid approach

internal/resources/attributes.go
```diff
@@ -28,6 +28,37 @@ const (
...
+	*schema.StringAttribute | *schema.Int64Attribute | *schema.BoolAttribute | *schema.SetAttribute |
+		*schema.Float64Attribute | *schema.NumberAttribute | *schema.ListAttribute
+}
+
+// WithDescription sets the MarkdownDescription for the provided attribute.
+// It accepts an attribute of type schema.StringAttribute, schema.Int64Attribute,
+// schema.BoolAttribute, schema.SetAttribute, schema.Float64Attribute,
+// schema.NumberAttribute, or schema.ListAttribute, and a description string.
+// The function returns the modified attribute.
```
itsHabib, line 41: can probably update the comment as well to only reference the type above

internal/resources/attributes.go
itsHabib: ya its obv not ideal but it does force a caller to only use the types we want which is good. maybe sometime in the future go will support field restriction on types like in typescript where u can say hey accept a type as an arg that has a certain field

internal/resources/attributes.go
PaulomeeCb: Done- https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/295/commits/7c407ac27dfccda538b03b2c331517f26f3ea520

internal/resources/attributes.go
PaulomeeCb: Done- https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/295/commits/cd8c131e3a7e54c507e7e37178fb1957e1202374

itsHabib (dismissed):

l0n3star (approved):

PR #360, opened 2025-07-02
[AV-105195] Add common functions for converting basetype strings and strings

mohammed-madi:
## Jira

* AV-105195

## Description

Add BaseStringsToStrings and StringsToBaseStrings to convert terraform basetype strings to generic strings and back
Add unit tests for both functions
Replace convertZones, MorphApiKeyOrganizationRoles and ConvertRoles functions which do the same thing but duplicate the code.

## Type of Change

- [ ] Manually tested
- [x] Unit tested
- [ ] Acceptance tested

### Testing

<details open>
  <summary>Testing</summary>
  
</details>

## Required Checklist:

## Further comments

### DIFF
--- a/internal/schema/apikey.go
+++ b/internal/schema/apikey.go
@@ -113,8 +113,7 @@ func MorphApiKeyResources(resources api.Resources) []ApiKeyResourcesItems {
 		if resource.Type != nil {
 			newResourceItem.Type = types.StringValue(*resource.Type)
 		}
-		newRoles := StringsToBaseStrings(resource.Roles)
-		newResourceItem.Roles = newRoles
+		newResourceItem.Roles = StringsToBaseStrings(resource.Roles)
 		newApiKeyResourcesItems = append(newApiKeyResourcesItems, newResourceItem)
 	}
 	return newApiKeyResourcesItems
--- a/internal/schema/apikey.go
+++ b/internal/schema/apikey.go
@@ -78,7 +78,7 @@ func NewApiKey(apiKey *api.GetApiKeyResponse, organizationId string, auditObject
 
 	newApiKey.AllowedCIDRs = newAllowedCidrs
 
-	newApiKey.OrganizationRoles = MorphApiKeyOrganizationRoles(apiKey.OrganizationRoles)
+	newApiKey.OrganizationRoles = StringsToBaseStrings(apiKey.OrganizationRoles)
 
 	newApiKey.Resources = MorphApiKeyResources(apiKey.Resources)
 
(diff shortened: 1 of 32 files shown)


internal/schema/user.go
```diff
@@ -112,17 +112,6 @@ func MorphRoles(roles []string) []basetypes.StringValue {
 	return morphedRoles
 }
 
-// ConvertRoles is used to convert all roles
-// in an array of basetypes.StringValue to strings.
-// TODO (AV-53457): add unit testing.
-func ConvertRoles(roles []basetypes.StringValue) []string {
```
PaulomeeCb, line 118: @mohammed-madi What about the other Convert functions here in this file- like `ConvertResources`? Do we need to replace them similarly?

internal/schema/apikey.go
```diff
@@ -101,17 +101,6 @@ func MorphAllowedCidrs(allowedCIDRs []string) (basetypes.SetValue, error) {
 	return newAllowedCidrs, nil
 }
 
-// MorphApiKeyOrganizationRoles is used to convert nested organizationRoles from
-// strings to terraform type.String.
-// TODO : add unit testing.
-func MorphApiKeyOrganizationRoles(organizationRoles []string) []basetypes.StringValue {
```
PaulomeeCb, line 107: What about the other functions like `MorphApiKeyResources`, do we need to replace that too similarly?

internal/schema/user.go
mohammed-madi: ConvertResources is more specific to API keys and should stay this way in my opinion, same for MorphApiKeyResources.   https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/360#discussion_r2180711618 
I've re-used the functions I added in a few different places around the API key code to simplify it.

internal/schema/types.go
```diff
@@ -0,0 +1,25 @@
...
+	result := make([]string, len(stringValues))
+	for i, sv := range stringValues {
+		result[i] = sv.ValueString()
+	}
+	return result
+}
+
+// StringsToBaseStrings converts a slice of strings to a slice of basetypes.StringValue.
+// It wraps each string in a basetypes.StringValue.
+func StringsToBaseStrings(strings []string) []basetypes.StringValue {
```
nidhi07kumar, line 19: small nit: Do not name the variable strings. It could conflict with package name.

internal/schema/types.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+)
+
+// BaseStringsToStrings converts a slice of basetypes.StringValue to a slice of strings.
+// It extracts the string values from each basetypes.StringValue element.
+func BaseStringsToStrings(stringValues []basetypes.StringValue) []string {
+	result := make([]string, len(stringValues))
+	for i, sv := range stringValues {
+		result[i] = sv.ValueString()
```
nidhi07kumar, line 12: What if the sv is `null` or `unknown` here? We should do a check before sv.ValueString().

internal/schema/types_test.go
l0n3star, line 1: nice to see unit tests !  

cc @Talina06

internal/schema/types_test.go
```diff
@@ -0,0 +1,141 @@
...
+			assert.Equal(t, len(test.expectedOutput), len(result))
+
+			for i, expected := range test.expectedOutput {
+				assert.Equal(t, expected.ValueString(), result[i].ValueString())
+			}
+		})
+	}
+}
+
+func TestConversionRoundTrip(t *testing.T) {
```
l0n3star, line 110: this seems redundant.    there are already tests for converting string to `basetypes.StringValue`, and `basetypes.StringValue` to string.

mohammed-madi: yeah, this was generated by cursor, will remove.

PR #366, opened 2025-07-30
[AV-107162] Do not allow setting autoexpansion for AWS/GCP

l0n3star:
## Jira

* AV-107162

## Description

`autoexpansion` is only valid for azure.  do not allow setting this for aws/gcp.

## Type of Change

- [X] Manually tested
- [ ] Unit tested
- [X] Acceptance tested

### Testing

<details open>
  <summary>Testing</summary>
  
validation in https://jira.issues.couchbase.com/browse/AV-107162?focusedCommentId=1239520

</details>

## Required Checklist:

## Further comments

### DIFF
--- a/internal/resources/cluster_schema.go
+++ b/internal/resources/cluster_schema.go
@@ -95,7 +95,7 @@ func ClusterSchema() schema.Schema {
 										"iops": WithDescription(int64Attribute(optional, computed),
 											"The number of IOPS for the disk. Only applicable for certain disk types."),
 										"autoexpansion": WithDescription(boolAttribute(optional, computed),
-											"Enable or disable automatic disk expansion."),
+											"Enable or disable automatic disk expansion.  This can only be set for Azure."),
 									},
 								},
 							},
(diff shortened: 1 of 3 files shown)


internal/resources/cluster_schema.go
```diff
@@ -95,7 +95,7 @@ func ClusterSchema() schema.Schema {
 										"iops": WithDescription(int64Attribute(optional, computed),
 											"The number of IOPS for the disk. Only applicable for certain disk types."),
 										"autoexpansion": WithDescription(boolAttribute(optional, computed),
-											"Enable or disable automatic disk expansion."),
+											"Enable or disable automatic disk expansion.  This can only be set for Azure."),
```
PaulomeeCb, line 98: I think similar to this we should update the descriptions for `iops` as well, it is only applicable for aws and azure right? Can you add a similar line to iops too?

PaulomeeCb (commented):
Thanks for the fix @l0n3star QQ. Are we doing a similar validation for iops as well?

internal/resources/cluster_schema.go
l0n3star: good catch.  this bug will also apply to iops.  will address in another PR.  filed https://jira.issues.couchbase.com/browse/AV-107267

PR #402, opened 2025-09-19
[AV-111247] Do not set bucket defaults in provider

l0n3star:
## Jira

* AV-111247

## Description

provider should not set bucket defaults.  this should be set by backend.

## Type of Change

- [X] Manually tested
- [ ] Unit tested
- [ ] Acceptance tested

### Testing

<details open>
  <summary>Testing</summary>
  
test validation in ticket

</details>

## Required Checklist:

## Further comments

### DIFF
--- a/internal/resources/bucket.go
+++ b/internal/resources/bucket.go
@@ -308,7 +308,9 @@ func (c *Bucket) ImportState(ctx context.Context, req resource.ImportStateReques
 }
 
 // retrieveBucket retrieves bucket information for a specified organization, project, cluster and bucket ID.
-func (c *Bucket) retrieveBucket(ctx context.Context, organizationId, projectId, clusterId, bucketId string) (*providerschema.OneBucket, error) {
+func (c *Bucket) retrieveBucket(
+	ctx context.Context, organizationId, projectId, clusterId, bucketId string,
+) (*providerschema.OneBucket, error) {
 	url := fmt.Sprintf("%s/v4/organizations/%s/projects/%s/clusters/%s/buckets/%s", c.HostURL, organizationId, projectId, clusterId, bucketId)
 	cfg := api.EndpointCfg{Url: url, Method: http.MethodGet, SuccessStatus: http.StatusOK}
 	response, err := c.Client.ExecuteWithRetry(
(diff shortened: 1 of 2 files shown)


internal/resources/bucket_schema.go
```diff
@@ -69,7 +64,6 @@ func BucketSchema() schema.Schema {
 			"time_to_live_in_seconds": schema.Int64Attribute{
 				Optional:            true,
 				Computed:            true,
-				Default:             int64default.StaticInt64(0),
```
Talina06, line 72: Can you add testing for these being removed as well?

Talina06 (changes requested):
need to see testing for all defaults that are removed

internal/resources/bucket_schema.go
l0n3star: please see tests here https://jira.issues.couchbase.com/browse/AV-111247?focusedCommentId=1268023

internal/resources/bucket.go
```diff
@@ -437,6 +439,28 @@ func initializeBucketWithPlanAndId(plan providerschema.Bucket, id string) provid
...
+	if plan.Flush.IsNull() || plan.Flush.IsUnknown() {
+		plan.Flush = types.BoolNull()
+	}
+	if plan.TimeToLiveInSeconds.IsNull() || plan.TimeToLiveInSeconds.IsUnknown() {
+		plan.TimeToLiveInSeconds = types.Int64Null()
+	}
+	if plan.Type.IsNull() || plan.Type.IsUnknown() {
+		plan.Type = types.StringNull()
+	}
+	// Stats is a read-only attribute, so we set it to null in the initial state
```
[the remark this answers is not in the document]
l0n3star: since defaults were removed, need to init computed values to null.  same as in other resources.

the reason is the framework requires all computed attributes have known value.  if refresh fails and we don't do this then terraform will show an error that there is a bug in provider.

there is probably a better way of doing this, but for now keeping implementation consistent across resources.

l0n3star:
> @l0n3star we also need to add acceptance tests for this change so we can capture any regressions.

this will take a lot of effort .  currently it only tests on latest Couchbase version.  we'll have to create multiple clusters.

PR #404, opened 2025-09-19
[AV-111366] Introduce a V2 API client via openapi-codegen

Talina06:
## Jira

* AV-111366

## Description

Introduces a generated OpenAPI v2 client alongside the existing v1, laying the groundwork for a safe, step-by-step migration.

### Improvements

1. Spec alignment - the client is generated directly from the V4 OpenAPI spec, ensuring consistency and eliminating manual drift.

2. Type safety - stronger typing helps catch issues earlier and makes code easier to navigate.

3. Maintainability - reproducible codegen pipeline replaces manual plumbing, making iteration and updates simpler.

4. Incorporates the retry logic from the previous client to keep behavior consistent.

### Solutoin

This PR wires the new v2 client into the provider while leaving existing resources on v1 for now. No migrations are included yet, the goal is to provide a solid foundation for incremental adoption in future PRs.

### Next steps
Follow-up changes will migrate specific resources (projects, buckets, clusters, etc.) to v2 and unify retries, logging, and error handling across clients.

## Type of Change

- [ ] Manually tested
- [x] Unit tested
- [ ] Acceptance tested

### Testing

<details open>
  <summary>Testing</summary>
  
</details>
Acceptance testing

## Required Checklist:

## Further comments


internal/generated/api/retry_http_client.go
```diff
@@ -0,0 +1,112 @@
...
+
+// defaultWaitAttempt re-attempt http request after 2 seconds.
+const defaultWaitAttempt = time.Second * 2
+
+func (t *RetryTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	// Buffer the body so we can retry non-idempotent requests as well
+	var bodyBytes []byte
+	if req.Body != nil {
+		b, _ := io.ReadAll(req.Body)
+		_ = req.Body.Close()
```
PaulomeeCb, line 35: here too..

internal/generated/api/retry_http_client.go
```diff
@@ -0,0 +1,112 @@
...
+					backoff = time.Second * time.Duration(secs)
+				}
+			}
+			// Drain and retry
+			_, _ = io.Copy(io.Discard, res.Body)
+			_ = res.Body.Close()
+			if err := sleepOrDone(req.Context(), backoff); err != nil {
+				return res, err
+			}
+			continue
```
PaulomeeCb, line 70: Because maybe these will go into infinite loop retrying forever without a limit

internal/generated/api/retry_http_client.go
```diff
@@ -0,0 +1,112 @@
...
+
+func (t *RetryTransport) base() http.RoundTripper {
+	if t.Base != nil {
+		return t.Base
+	}
+	return http.DefaultTransport
+}
+
+// defaultWaitAttempt re-attempt http request after 2 seconds.
+const defaultWaitAttempt = time.Second * 2
```
PaulomeeCb, line 28: Nit: an improvement can be to implement exponential backoff.

PaulomeeCb (approved):
Some comments on the retrying part, otherwise the changes lgtm for the given scope of this PR

internal/generated/api/retry_http_client.go
```diff
@@ -0,0 +1,332 @@
...
+	if delay > float64(maxBackoffDelay) {
+		delay = float64(maxBackoffDelay)
+	}
+
+	// Add jitter: +/-25% of the calculated delay to prevent thundering herd
+	jitter := delay * 0.25 * (rand.Float64() - 0.5) * 2
+	finalDelay := delay + jitter
+
+	// Ensure delay is not negative
+	if finalDelay < 0 {
```
l0n3star, line 112: why not use `uint` ?

internal/generated/api/retry_http_client.go
```diff
@@ -0,0 +1,332 @@
...
+//   - Attempt 4: ~16s (12s - 20s with jitter)
+//   - Attempt 5+: ~30s (22.5s - 37.5s with jitter, capped at maxBackoffDelay)
+func calculateBackoff(attempt int) time.Duration {
+	// Calculate exponential backoff: baseDelay * 2^attempt
+	delay := float64(baseBackoffDelay) * math.Pow(2, float64(attempt))
+
+	// Cap the delay at maxBackoffDelay
+	if delay > float64(maxBackoffDelay) {
+		delay = float64(maxBackoffDelay)
+	}
```
l0n3star, line 105: nit

can do 

`delay = min(delay, maxBackoffDelay)`

internal/generated/api/retry_http_client.go
```diff
@@ -0,0 +1,332 @@
...
+// Examples:
+//   - Attempt 0: ~1s (750ms - 1.25s with jitter)
+//   - Attempt 1: ~2s (1.5s - 2.5s with jitter)
+//   - Attempt 2: ~4s (3s - 5s with jitter)
+//   - Attempt 3: ~8s (6s - 10s with jitter)
+//   - Attempt 4: ~16s (12s - 20s with jitter)
+//   - Attempt 5+: ~30s (22.5s - 37.5s with jitter, capped at maxBackoffDelay)
+func calculateBackoff(attempt int) time.Duration {
+	// Calculate exponential backoff: baseDelay * 2^attempt
+	delay := float64(baseBackoffDelay) * math.Pow(2, float64(attempt))
```
l0n3star, line 100: no need to convert to float for 2^x

can use bit shift ie 2 << x

internal/generated/api/retry_http_client.go
```diff
@@ -0,0 +1,332 @@
+// Package api provides HTTP client functionality with intelligent retry logic
+// for the Couchbase Capella API. This package includes exponential backoff
+// retry mechanisms for handling rate limits and temporary server errors.
+package api
+
+import (
```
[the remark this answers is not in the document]
Talina06: I need to understand the lib before using, the plain way is preferred for now

internal/generated/api/retry_http_client.go
```diff
@@ -0,0 +1,332 @@
...
+
+	// Ensure delay is not negative
+	if finalDelay < 0 {
+		finalDelay = float64(baseBackoffDelay)
+	}
+
+	return time.Duration(finalDelay)
+}
+
+// RoundTrip executes a single HTTP transaction with intelligent retry logic.
```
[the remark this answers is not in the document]
Talina06: right, but it is not exported to be common, so I set it up separately for the v2 client, and we will deprecate v1 client from this point - only use it for existing resources until ported

internal/generated/api/retry_http_client.go
Talina06: what is min? another user defined function? why would we want to define it, if it's used only at one place?

internal/generated/api/retry_http_client.go
Talina06: jitter calculations can make it negative, and the formula is non-int

internal/generated/api/retry_http_client.go
```diff
@@ -0,0 +1,331 @@
...
+			var apiErr struct {
+				Code int `json:"code"`
+			}
+			if err := json.Unmarshal(b, &apiErr); err != nil {
+				// If we can't parse the error, treat it as a regular 504 and retry.
+				apiErr.Code = 0
+			}
+			if apiErr.Code == 7001 {
+				return res, nil
+			}
```
l0n3star, line 245: error should be returned here otherwise it causes regression.  see previous client implementation:

https://github.com/couchbasecloud/terraform-provider-couchbase-capella/blob/main/internal/resources/gsi.go#L238

internal/generated/api/retry_http_client.go
```diff
@@ -0,0 +1,331 @@
...
+
+			// Use exponential backoff for 504 errors.
+			backoff := calculateBackoff(retryCount)
+			if err := sleepOrDone(req.Context(), backoff); err != nil {
+				return res, err
+			}
+			retryCount++
+			continue
+		default:
+			return res, nil
```
l0n3star, line 255: maybe I'm missing something, but what happens when we get non-2xx response code ?

internal/generated/api/retry_http_client.go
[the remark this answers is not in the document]
Talina06: added

PaulomeeCb (dismissed):

internal/generated/api/retry_http_client.go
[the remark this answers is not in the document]
l0n3star: brilliant thanks

l0n3star (dismissed):

internal/generated/api/retry_http_client.go
```diff
@@ -0,0 +1,203 @@
...
+//     which indicates a non-retryable error condition for index DDL operations
+//   - All other status codes: Returns immediately without retry
+//
+// For 504 responses with code 7001, it returns the specific ErrGatewayTimeoutForIndexDDL
+// error to provide better error context to callers, matching V1 client behavior.
+//
+//nolint:nilerr // returning nil error is expected behavior for retryablehttp.CheckRetry interface
+func customRetryPolicy(ctx context.Context, resp *http.Response, err error) (bool, error) {
+	// Handle connection errors - retry these
+	if err != nil {
```
nidhi07kumar, line 57: @Talina06 Do we want to retry all errors? Can we skip any of them?

internal/generated/api/retry_http_client.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	switch resp.StatusCode {
+	case http.StatusTooManyRequests:
+		// Always retry 429 responses - retryablehttp handles Retry-After header automatically
+		return true, nil
+
+	case http.StatusGatewayTimeout:
+		// Check for special code 7001 (do not retry)
+		body, readErr := io.ReadAll(resp.Body)
```
nidhi07kumar, line 68: What if this is a large response? Do we want to read the full response body?

nidhi07kumar: should we just unmarshal here instead of reading the response?

Talina06: fixed in [c3348ea](https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/404/commits/c3348ea650cd02a6ab17ff519377d38df1b5379a)

nidhi07kumar (approved):
LGTM

l0n3star (dismissed):

nidhi07kumar (approved):

internal/generated/api/retry_http_client.go
```diff
@@ -0,0 +1,193 @@
...
+	case http.StatusTooManyRequests:
+		// Always retry 429 responses - retryablehttp handles Retry-After header automatically
+		return true, nil
+
+	case http.StatusGatewayTimeout:
+		// Check for special code 7001 (do not retry)
+		var apiErr struct {
+			Code int `json:"code"`
+		}
+		if err := json.NewDecoder(resp.Body).Decode(&apiErr); err != nil {
```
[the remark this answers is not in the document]
PaulomeeCb: Okay looks like the retry mechanism is working correctly. Thanks for adding the unit tests 👍 What I mentioned above was more about response body consumption, not about the retry decision logic(In case someone tries to read the current response, after consumption, it will be empty)

PaulomeeCb (approved):

PR #544, opened 2026-04-07
[AV-125341] Add private endpoint DNS 

l0n3star:
## Jira

* AV-125341

## Description

expose the Private Endpoint URL in private endpoint resource

## Type of Change

- [X] Manually tested
- [ ] Unit tested
- [ ] Acceptance tested

### Testing

<details open>
  <summary>Testing</summary>
  
validation in ticket

</details>

## Further comments


matty271828 (commented):
Should there be an additional datasource for this? Or is that the same as already implemented in https://github.com/couchbasecloud/terraform-provider-couchbase-capella/blob/e1a1d9680ceda90d7a9ca3ab3ae35fe64d2d4e47/internal/datasources/private_endpoints.go ?

.factory/skills/tf-examples-gen/SKILL.md
```diff
@@ -0,0 +1,103 @@
+---
```
matty271828, line 1: See comments on https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/547

Lackshan (dismissed):
This is what I had: https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/549/changes, so this looks good to me.

l0n3star:
> Should there be an additional datasource for this? Or is that the same as already implemented in https://github.com/couchbasecloud/terraform-provider-couchbase-capella/blob/e1a1d9680ceda90d7a9ca3ab3ae35fe64d2d4e47/internal/datasources/private_endpoints.go ?

per the PRD just need a resource

toshikagarwal:
> > Should there be an additional datasource for this? Or is that the same as already implemented in https://github.com/couchbasecloud/terraform-provider-couchbase-capella/blob/e1a1d9680ceda90d7a9ca3ab3ae35fe64d2d4e47/internal/datasources/private_endpoints.go ?
> 
> per the PRD just need a resource

@niranjs1  FYI

niranjs1:
> > Should there be an additional datasource for this? Or is that the same as already implemented in https://github.com/couchbasecloud/terraform-provider-couchbase-capella/blob/e1a1d9680ceda90d7a9ca3ab3ae35fe64d2d4e47/internal/datasources/private_endpoints.go ?
> 
> per the PRD, just need a resource

Was thinking that the user would ever want this value only while owning the resource in Terraform state ( because this is part of the flow to provision and configuration of private endpoints ), hence just the resource.  What's the best practice here?  If updating the datasource, would prefer updating the existing one.

l0n3star:
> > > Should there be an additional datasource for this? Or is that the same as already implemented in https://github.com/couchbasecloud/terraform-provider-couchbase-capella/blob/e1a1d9680ceda90d7a9ca3ab3ae35fe64d2d4e47/internal/datasources/private_endpoints.go ?
> > 
> > 
> > per the PRD, just need a resource
> 
> Was thinking that the user would ever want this value only while owning the resource in Terraform state ( because this is part of the flow to provision and configuration of private endpoints ), hence just the resource. What's the best practice here? If updating the datasource, would prefer updating the existing one.

datasource might be useful if you want to query a resource you don't own.

will update existing datasource.

niranjs1:
> > > > Should there be an additional datasource for this? Or is that the same as already implemented in https://github.com/couchbasecloud/terraform-provider-couchbase-capella/blob/e1a1d9680ceda90d7a9ca3ab3ae35fe64d2d4e47/internal/datasources/private_endpoints.go ?
> > > 
> > > 
> > > per the PRD, just need a resource
> > 
> > 
> > Was thinking that the user would ever want this value only while owning the resource in Terraform state ( because this is part of the flow to provision and configuration of private endpoints ), hence just the resource. What's the best practice here? If updating the datasource, would prefer updating the existing one.
> 
> datasource might be useful if you want to query a resource you don't own
Makes sens.  Let's update the existing datasource to include this value @Lackshan . Thanks!

matty271828 (commented):
Code looks good. See we have acceptance test failure though and can we also add an acceptance test for this please?

l0n3star:
> Code looks good. See we have acceptance test failure though and can we also add an acceptance test for this please?

will do that separately.  requires adding a private endpoint in a "customer" VPC so will take some effort.

internal/resources/private_endpoints.go
```diff
@@ -298,20 +292,56 @@ func (p *PrivateEndpoint) getPrivateEndpointStatus(ctx context.Context, organiza
...
+			)
+			if err != nil {
+				return err
+			}
+
+			if status == "linked" && dns != "" {
+				plan.ServiceName = types.StringValue(serviceName)
+				plan.Status = types.StringValue(status)
+				plan.PrivateEndpointDNS = types.StringValue(dns)
+				return nil
```
toshikagarwal, line 342: Reviewed the code using Factory and got a warning about 
`Timer Leak in waitUntilLinked (Medium) The timer is not stopped when returning from the function, which could cause a goroutine leak`
Suggested:
```
timer := time.NewTimer(time.Second * 1)
defer timer.Stop()
```

l0n3star: addressed in https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/544/commits/62e3e844c7a9127a24ae3588b32b65ac6e3af8a5

PaulomeeCb (commented):
@droid

internal/resources/private_endpoints.go
```diff
@@ -298,20 +292,57 @@ func (p *PrivateEndpoint) getPrivateEndpointStatus(ctx context.Context, organiza
...
+			}
```
nidhi07kumar: If dns == "", we keep retrying silently until timeout. Can you add a debug log here when we are waiting for it to be linked? 

```suggestion
			if status == "linked" {
			     dns == "" {
			     // debug log
			     timer.Reset(time.minute *1)
			     continue
			     }
			
				plan.ServiceName = types.StringValue(serviceName)
				plan.Status = types.StringValue(status)
				plan.PrivateEndpointDNS = types.StringValue(dns)
				return nil
			}
```

nidhi07kumar: Also, declare "linked" as a const.

l0n3star: addressed in https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/544/commits/b344e4fe4f4304c600d52a5a7d252fadfc850c22

nidhi07kumar (approved):
LGTM

PR #550, opened 2026-04-09
Fix copy_to_regions optional field causing Value Conversion Error

TJM:
## Summary

- `copy_to_regions` was typed as `[]types.String` in the schema struct but declared as a `SetAttribute` in the resource schema. When the field is omitted from config, the Terraform framework sets it to an unknown value that a plain Go slice cannot represent — only `types.Set` can.
- Changed `CopyToRegions` field from `[]types.String` to `types.Set` and updated all conversion logic accordingly.
- Added an acceptance test covering creation without `copy_to_regions` set.

## Reproduction

```hcl
resource "couchbase-capella_cloud_snapshot_backup_schedule" "example" {
  interval        = 4
  retention       = 168
  start_time      = "2024-01-01T00:00:00+00:00"
  cluster_id      = couchbase-capella_cluster.example.id
  project_id      = couchbase-capella_cluster.example.project_id
  organization_id = local.capella_org_id
}
```

Previously produced:
```
│ Error: Value Conversion Error
│ Path: copy_to_regions
│ Target Type: []basetypes.StringValue
│ Suggested Type: basetypes.SetValue
```

## Test plan

- [ ] `TestAccSnapshotBackupScheduleResourceWithoutCopyToRegions` — new test, creates resource without `copy_to_regions`
- [ ] `TestAccSnapshotBackupScheduleResource` — existing happy-path test still passes
- [ ] `TestAccSnapshotBackupScheduleResourceInvalidCopyToRegions` — existing negative test still passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/acceptance_tests/snapshot_backup_schedule_acceptance_test.go
+++ b/acceptance_tests/snapshot_backup_schedule_acceptance_test.go
@@ -132,6 +132,46 @@ func TestAccSnapshotBackupScheduleResourceInvalidCopyToRegions(t *testing.T) {
 	})
 }
 
+func TestAccSnapshotBackupScheduleResourceWithoutCopyToRegions(t *testing.T) {
+	resourceName := randomStringWithPrefix("tf_acc_cloud_snapshot_backup_schedule_")
+	resourceReference := "couchbase-capella_cloud_snapshot_backup_schedule." + resourceName
+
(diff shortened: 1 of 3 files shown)


TJM:
A workaround for anyone trying to set these up before this MR is merged, you can set copy_to_regions to an empty set:

```hcl
  copy_to_regions = toset([])
```

PR #597, opened 2026-05-18
[AV-131315] Fix racy getLatestBackup that hangs couchbase-capella_backup Create up to 90 min

panigrahisubhrajit:
## Jira

* [AV-131315](https://jira.issues.couchbase.com/browse/AV-131315) — `couchbase-capella_backup` Create hangs up to 90 min on busy clusters due to racy `getLatestBackup`

### Problem

`couchbase-capella_backup` Create can hang for up to 90 minutes during `terraform apply`. In CI this exhausts the 2-hour Go test-binary budget and panics the whole suite ([Acceptance Tests run 25947971975](https://github.com/couchbasecloud/terraform-provider-couchbase-capella/actions/runs/25947971975): `TestAccBackupResource (1h24m29s)` still running when the 2h panic fired).

### Root cause

`getLatestBackup` in [`internal/resources/backup.go`](internal/resources/backup.go) does not actually return the **latest** backup. It calls `GET /clusters/{id}/backups` with:

* no `sortBy`,
* no pagination (no `cursor` field on the response struct, so only page 1 is read),

then returns the first record on page 1 whose `bucketID` matches. On busy clusters that's often a stale historical record — not the one Create just submitted.

`checkLatestBackupStatus` then loops waiting for an id different from that stale one, in a final state. The stale record stays pinned, the new POST's record sits on a later page, and the loop burns the full 90-minute wait window every time.

### Fix

| File | Change |
|---|---|
| [`internal/resources/backup.go`](internal/resources/backup.go) — `getLatestBackup` | Walk every page via `api.GetPaginated[]`, filter to the target bucket client-side, and pick the entry with the max `Date`. "Latest" is now actually the latest regardless of the API's default ordering or how many historical records the cluster has accumulated. |
| [`internal/resources/backup.go`](internal/resources/backup.go) — `checkLatestBackupStatus` | Reduce wait timeout from **90 min → 30 min** so one stuck backup can no longer consume more than a quarter of the 120-min Go test-binary budget. |
| [`internal/api/client.go`](internal/api/client.go) — `exec` | Cap `ExecuteWithRetry` at **10 total attempts** (initial + up to 9 retries) on 429/504. Pathological 429-flap can no longer silently burn the 10-minute outer envelope. Also fix a nil-wrap bug on `ctx.Done()` so a context timeout no longer returns `nil` error to callers. |
| [`acceptance_tests/backup_acceptance_test.go`](acceptance_tests/backup_acceptance_test.go) — `TestAccBackupResource` | Run on a fresh dedicated bucket created per test run rather than the shared `globalBucketId`. On the CI tenant a leaked pending backup from a previously-killed run can queue this test's POST indefinitely (legacy bucket-backup is serialised per bucket); a dedicated bucket guarantees zero accumulated state. |

## Type of Change

- [ ] Manually tested
- [ ] Unit tested
- [x] Acceptance tested

### Testing

<details open>
  <summary>Testing</summary>

* `go build ./...` clean
* `go vet ./...` clean
* `go test -c ./acceptance_tests` clean
* `TestAccBackupResource` PASSED in **95.42 s** on the busy CI tenant — compared to the 1h24m+ hang on the same tenant before the fix.
* All other `TestAccBackup*` tests pass.

</details>

## Further comments

**Acceptance criteria from [AV-131315](https://jira.issues.couchbase.com/browse/AV-131315):**

* [x] `getLatestBackup` walks all pages and sorts client-side by date desc
* [x] Create poll timeout lowered from 90 min → 30 min
* [x] `ExecuteWithRetry` capped at 10 attempts
* [x] `TestAccBackupResource` completes in under 5 min on the busy CI tenant (actual: ~95 s)
* [x] No `panic: test timed out` on the test binary

**Unrelated CI failures still visible on this PR's Build:** `TestAccBackupScheduleResourceInvalidType` (fixed in PR #599) and the `TestAccAppServiceLogStreaming/App_Endpoint_Logging_Config` HTTP-timeout flake — neither relates to this PR's scope.

[AV-131315]: https://couchbasecloud.atlassian.net/browse/AV-131315?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AV-131315]: https://couchbasecloud.atlassian.net/browse/AV-131315?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/api/client.go
+++ b/internal/api/client.go
@@ -57,9 +57,8 @@ type EndpointCfg struct {
 // defaultWaitAttempt re-attempt http request after 2 seconds.
 const defaultWaitAttempt = time.Second * 2
 
-// maxRetryAttempts caps 429/504 retries independently of the 10-minute wall-clock
-// timeout, so a rapid flap with small backoffs cannot exhaust the full window.
-const maxRetryAttempts = 10
+// maxAttempts caps total 429/504 attempts independently of the 10-minute wall-clock timeout.
+const maxAttempts = 10
 
 
 // ExecuteWithRetry is used to construct and execute a HTTP request with retry.
@@ -199,8 +198,8 @@ func exec(
 				return response, err
 			}
 
-			if attempts >= maxRetryAttempts {
-				return nil, fmt.Errorf("exhausted %d retry attempts: %w", attempts, err)
+			if attempts >= maxAttempts {
+				return nil, fmt.Errorf("exhausted %d attempts: %w", attempts, err)
 			}
 
 			if backOff > 0 {
--- a/internal/api/client.go
+++ b/internal/api/client.go
@@ -57,13 +57,8 @@ type EndpointCfg struct {
 // defaultWaitAttempt re-attempt http request after 2 seconds.
 const defaultWaitAttempt = time.Second * 2
 
-// maxRetryAttempts caps the total number of attempts made by ExecuteWithRetry
-// on 429/504, including the initial request. With a value of 10, the client
-// will make up to 9 retries after the first failed attempt. Combined with the
-// 10-minute outer envelope below, this stops a pathological flap from silently
-// burning the test-binary timeout budget. The wall-clock bound is governed by
-// the 10-minute outer timeout — a 429 response honors the server's Retry-After
-// header, which can itself be large.
+// maxRetryAttempts caps 429/504 retries independently of the 10-minute wall-clock
+// timeout, so a rapid flap with small backoffs cannot exhaust the full window.
 const maxRetryAttempts = 10
 
 // Execute is used to construct and execute a HTTP request.
--- a/internal/api/client.go
+++ b/internal/api/client.go
@@ -244,7 +244,7 @@ func exec(
 	for {
 		select {
 		case <-ctx.Done():
-			return nil, fmt.Errorf("timed out executing request against api after %d attempts: %w", attempts, err)
+			return nil, fmt.Errorf("timed out executing request against api after %d attempts: %w", attempts, ctx.Err())
 		case <-timer.C:
 			response, backOff, err = fn()
 			attempts++
--- a/internal/api/client.go
+++ b/internal/api/client.go
@@ -61,8 +61,9 @@ const defaultWaitAttempt = time.Second * 2
 // on 429/504, including the initial request. With a value of 10, the client
 // will make up to 9 retries after the first failed attempt. Combined with the
 // 10-minute outer envelope below, this stops a pathological flap from silently
-// burning the test-binary timeout budget. Picked so the worst case stays under
-// ~2 min even when each attempt completes quickly.
+// burning the test-binary timeout budget. The wall-clock bound is governed by
+// the 10-minute outer timeout — a 429 response honors the server's Retry-After
+// header, which can itself be large.
 const maxRetryAttempts = 10
 
 // Execute is used to construct and execute a HTTP request.
--- a/internal/api/client.go
+++ b/internal/api/client.go
@@ -57,6 +57,12 @@ type EndpointCfg struct {
 // defaultWaitAttempt re-attempt http request after 2 seconds.
 const defaultWaitAttempt = time.Second * 2
 
+// maxRetryAttempts caps the number of times ExecuteWithRetry will resend a request
+// on 429/504. Combined with the 10-minute outer envelope below, this stops a
+// pathological flap from silently burning the test-binary timeout budget. Picked
+// so the worst case stays under ~2 min even when each attempt completes quickly.
+const maxRetryAttempts = 10
+
 // Execute is used to construct and execute a HTTP request.
 // It then returns the response.
 func (c *Client) Execute(
@@ -218,11 +224,13 @@ func exec(
 	ctx context.Context, fn func() (response *Response, dur time.Duration, err error), waitOnReattempt time.Duration,
 ) (*Response, error) {
 	timer := time.NewTimer(time.Millisecond)
+	defer timer.Stop()
 
 	var (
 		err      error
 		backOff  time.Duration
 		response *Response
+		attempts int
 	)
 
 	const timeout = time.Minute * 10
@@ -234,15 +242,20 @@ func exec(
 	for {
 		select {
 		case <-ctx.Done():
-			return nil, fmt.Errorf("timed out executing request against api: %w", err)
+			return nil, fmt.Errorf("timed out executing request against api after %d attempts: %w", attempts, err)
 		case <-timer.C:
 			response, backOff, err = fn()
+			attempts++
 			switch {
 			case err == nil:
 				return response, nil
 			case goer.Is(err, errors.ErrRatelimit):
 			case !goer.Is(err, errors.ErrGatewayTimeout):
 				return response, err
+			}
+
+			if attempts >= maxRetryAttempts {
+				return nil, fmt.Errorf("exhausted %d retry attempts: %w", attempts, err)
 			}
 
 			if backOff > 0 {
--- a/internal/api/client.go
+++ b/internal/api/client.go
@@ -57,10 +57,6 @@ type EndpointCfg struct {
 // defaultWaitAttempt re-attempt http request after 2 seconds.
 const defaultWaitAttempt = time.Second * 2
 
-// maxAttempts caps total 429/504 attempts independently of the 10-minute wall-clock timeout.
-const maxAttempts = 10
-
-
 // ExecuteWithRetry is used to construct and execute a HTTP request with retry.
 // It then returns the response.
 func (c *Client) ExecuteWithRetry(
@@ -174,7 +170,6 @@ func exec(
 		err      error
 		backOff  time.Duration
 		response *Response
-		attempts int
 	)
 
 	const timeout = time.Minute * 10
@@ -186,20 +181,15 @@ func exec(
 	for {
 		select {
 		case <-ctx.Done():
-			return nil, fmt.Errorf("timed out executing request against api after %d attempts: %w", attempts, ctx.Err())
+			return nil, fmt.Errorf("timed out executing request against api: %w", ctx.Err())
 		case <-timer.C:
 			response, backOff, err = fn()
-			attempts++
 			switch {
 			case err == nil:
 				return response, nil
 			case goer.Is(err, errors.ErrRatelimit):
 			case !goer.Is(err, errors.ErrGatewayTimeout):
 				return response, err
-			}
-
-			if attempts >= maxAttempts {
-				return nil, fmt.Errorf("exhausted %d attempts: %w", attempts, err)
 			}
 
 			if backOff > 0 {
--- a/internal/api/client.go
+++ b/internal/api/client.go
@@ -61,59 +61,6 @@ const defaultWaitAttempt = time.Second * 2
 // timeout, so a rapid flap with small backoffs cannot exhaust the full window.
 const maxRetryAttempts = 10
 
-// Execute is used to construct and execute a HTTP request.
-// It then returns the response.
-func (c *Client) Execute(
-	endpointCfg EndpointCfg,
-	payload any,
-	authToken string,
-	headers map[string]string,
-) (response *Response, err error) {
-	var requestBody []byte
-	if payload != nil {
-		requestBody, err = json.Marshal(payload)
-		if err != nil {
-			return nil, fmt.Errorf("%s: %w", errors.ErrMarshallingPayload, err)
-		}
-	}
-
-	req, err := http.NewRequest(endpointCfg.Method, endpointCfg.Url, bytes.NewReader(requestBody))
-	if err != nil {
-		return nil, fmt.Errorf("%s: %w", errors.ErrConstructingRequest, err)
-	}
-
-	req.Header.Set("Authorization", "Bearer "+authToken)
-	req.Header.Set("User-Agent", userAgent)
-	for header, value := range headers {
-		req.Header.Set(header, value)
-	}
-
-	apiRes, err := c.Do(req)
-	if err != nil {
-		return nil, fmt.Errorf("%s: %w", errors.ErrExecutingRequest, err)
-	}
-	defer apiRes.Body.Close()
-
-	responseBody, err := io.ReadAll(apiRes.Body)
-	if err != nil {
-		return
-	}
-
-	if apiRes.StatusCode != endpointCfg.SuccessStatus {
-		var apiError Error
-		if err := json.Unmarshal(responseBody, &apiError); err != nil {
-			return nil, fmt.Errorf(
-				"unexpected code: %d, expected: %d, body: %s",
-				apiRes.StatusCode, endpointCfg.SuccessStatus, responseBody)
-		}
-		return nil, &apiError
-	}
-
-	return &Response{
-		Response: apiRes,
-		Body:     responseBody,
-	}, nil
-}
 
 // ExecuteWithRetry is used to construct and execute a HTTP request with retry.
 // It then returns the response.
--- a/internal/api/client.go
+++ b/internal/api/client.go
@@ -57,10 +57,12 @@ type EndpointCfg struct {
 // defaultWaitAttempt re-attempt http request after 2 seconds.
 const defaultWaitAttempt = time.Second * 2
 
-// maxRetryAttempts caps the number of times ExecuteWithRetry will resend a request
-// on 429/504. Combined with the 10-minute outer envelope below, this stops a
-// pathological flap from silently burning the test-binary timeout budget. Picked
-// so the worst case stays under ~2 min even when each attempt completes quickly.
+// maxRetryAttempts caps the total number of attempts made by ExecuteWithRetry
+// on 429/504, including the initial request. With a value of 10, the client
+// will make up to 9 retries after the first failed attempt. Combined with the
+// 10-minute outer envelope below, this stops a pathological flap from silently
+// burning the test-binary timeout budget. Picked so the worst case stays under
+// ~2 min even when each attempt completes quickly.
 const maxRetryAttempts = 10
 
 // Execute is used to construct and execute a HTTP request.
--- a/internal/resources/backup.go
+++ b/internal/resources/backup.go
@@ -75,11 +75,6 @@ func (b *Backup) Create(ctx context.Context, req resource.CreateRequest, resp *r
 	var clusterId = plan.ClusterId.ValueString()
 	var bucketId = plan.BucketId.ValueString()
 
-	// Capture the bucket's current latest backup BEFORE the POST so we can detect
-	// when the new record appears. getLatestBackup correctly walks all pages and
-	// sorts client-side by date; the only thing it cannot do is identify "our"
-	// record at POST time (Capella's backup POST returns 202 with an empty body,
-	// unlike cluster/appservice POSTs which echo the new id).
 	latestBackup, err := b.getLatestBackup(ctx, organizationId, projectId, clusterId, bucketId)
 	if err != nil {
 		resp.Diagnostics.AddError(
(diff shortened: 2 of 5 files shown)


panigrahisubhrajit:
PR  functionally solves the ticket. The Build failure is a different issue , it doesn't relate to the changes in the PR. 95-second runtime confirms TestAccBackupResource no longer hangs the 2-hour test budget on busy clusters.

internal/resources/backup.go
```diff
@@ -492,33 +496,36 @@ func (b *Backup) retrieveBackup(ctx context.Context, organizationId, projectId,
 	return refreshedState, nil
 }
 
-// getLatestBackup retrieves the latest backup information for a specified bucket in a cluster
-// from the specified organization, project and cluster using the provided bucket ID by open-api call.
+// getLatestBackup returns the most-recent backup record for the given bucket
+// in a cluster, or (nil, nil) if the bucket has no backups yet.
+//
+// The prior implementation called the cluster-scoped list endpoint without
```
SophieWegmann: Nit - I'm not sure if we need a section about the previous implementation, I think just explaining the reasoning behind the current implementation should be enough

SophieWegmann (commented):
Looks good - just one nit

internal/api/client.go
```diff
@@ -57,6 +57,15 @@ type EndpointCfg struct {
...
+// will make up to 9 retries after the first failed attempt. Combined with the
+// 10-minute outer envelope below, this stops a pathological flap from silently
+// burning the test-binary timeout budget. The wall-clock bound is governed by
+// the 10-minute outer timeout — a 429 response honors the server's Retry-After
+// header, which can itself be large.
+const maxRetryAttempts = 10
+
 // Execute is used to construct and execute a HTTP request.
 // It then returns the response.
 func (c *Client) Execute(
```
Lagher0: I don't see this execute being used anywhere in the codebase. Why do we have it?

internal/resources/backup.go
```diff
@@ -402,53 +407,52 @@ func (a *Backup) validateCreateBackupRequest(plan providerschema.Backup) error {
...
+// backup can no longer consume more than a quarter of the 120-minute Go
+// test-binary budget. Real-cluster backups of very large buckets that need
+// longer than 30 minutes will require widening this constant — the resource
+// schema does not currently expose a per-resource timeouts block.
 func (b *Backup) checkLatestBackupStatus(ctx context.Context, organizationId, projectId, clusterId, bucketId string, backupFound bool, latestBackup *backupapi.GetBackupResponse) (*backupapi.GetBackupResponse, error) {
-	var (
-		backupResp *backupapi.GetBackupResponse
-		err        error
+	const (
+		timeout = 30 * time.Minute
```
IsaacLambat: This shortening of the timeout may fix the acceptance test from timing out, but surely this could cause actual customer issues. The function comment even mentions this.

We shouldn't be causing issues for customers to make a test pass reliably. We should find a different want to do this.

internal/api/client.go
```diff
@@ -245,6 +257,10 @@ func exec(
 				return response, err
 			}
 
+			if attempts >= maxRetryAttempts {
```
IsaacLambat: Not against this, but I would have thought the `timeout` makes this not needed?

internal/resources/backup.go
IsaacLambat: Agreed. Seems to be a common theme with AI generated code comments in that it writes lots of long comments with a lot of information that can either be easily inferred from readable code, or just details of past changes that are not really needed.

internal/resources/backup.go
```diff
@@ -75,6 +75,11 @@ func (b *Backup) Create(ctx context.Context, req resource.CreateRequest, resp *r
 	var clusterId = plan.ClusterId.ValueString()
 	var bucketId = plan.BucketId.ValueString()
 
+	// Capture the bucket's current latest backup BEFORE the POST so we can detect
+	// when the new record appears. getLatestBackup correctly walks all pages and
+	// sorts client-side by date; the only thing it cannot do is identify "our"
+	// record at POST time (Capella's backup POST returns 202 with an empty body,
+	// unlike cluster/appservice POSTs which echo the new id).
```
Lagher0: Why do we need this comment here?

internal/resources/backup.go
panigrahisubhrajit: Primary fix is something else, timeout was reduced as a secondary safety net. I didnt want the test to just hang there. Anyway fixed it now.

internal/api/client.go
panigrahisubhrajit: here 10-minute timeout is the primary bound, but maxRetryAttempts short-circuits faster when backoffs are small. 
if you want to bail earlier than 10 minutes on rapid 429 flaps (for example stop after 10 attempts at 2s each, exit in 20s instead of running the full 10 minutes)

internal/resources/backup.go
IsaacLambat: Ok sounds good, what is the primary fix?

Guessing it is the `GetPaginated` use and dedicated bucket?

panigrahisubhrajit: yes thats the fix. Sometimes it can take more time than expected.

internal/api/client.go
```diff
@@ -57,6 +57,10 @@ type EndpointCfg struct {
 // defaultWaitAttempt re-attempt http request after 2 seconds.
 const defaultWaitAttempt = time.Second * 2
 
+// maxRetryAttempts caps 429/504 retries independently of the 10-minute wall-clock
+// timeout, so a rapid flap with small backoffs cannot exhaust the full window.
+const maxRetryAttempts = 10
```
Lagher0: Why are we using 10 attempts?

internal/api/client.go
```diff
@@ -57,6 +57,10 @@ type EndpointCfg struct {
 // defaultWaitAttempt re-attempt http request after 2 seconds.
 const defaultWaitAttempt = time.Second * 2
 
+// maxRetryAttempts caps 429/504 retries independently of the 10-minute wall-clock
+// timeout, so a rapid flap with small backoffs cannot exhaust the full window.
```
Lagher0: Shouldn't we have exponential backoffs, so that we don't burn through the small backoffs too quickly and fail the test?

internal/api/client.go
Lagher0, line 163: How will adding the retry logic affect the rest of the tests?

internal/api/client.go
```diff
@@ -57,6 +57,10 @@ type EndpointCfg struct {
...
+const maxRetryAttempts = 10
+
 // Execute is used to construct and execute a HTTP request.
 // It then returns the response.
 func (c *Client) Execute(
```
Lagher0: I don't see this function used anywhere in the codebase. Why do we have it?

internal/resources/backup.go
```diff
@@ -492,33 +487,26 @@ func (b *Backup) retrieveBackup(ctx context.Context, organizationId, projectId,
...
 
-// getLatestBackup retrieves the latest backup information for a specified bucket in a cluster
-// from the specified organization, project and cluster using the provided bucket ID by open-api call.
+// getLatestBackup returns the most-recent backup record for the given bucket
+// in a cluster, or (nil, nil) if the bucket has no backups yet.
+// It walks all pages, filters to the target bucket, and sorts by Date desc.
 func (b *Backup) getLatestBackup(ctx context.Context, organizationId, projectId, clusterId, bucketId string) (*backupapi.GetBackupResponse, error) {
 	url := fmt.Sprintf("%s/v4/organizations/%s/projects/%s/clusters/%s/backups", b.HostURL, organizationId, projectId, clusterId)
 	cfg := api.EndpointCfg{Url: url, Method: http.MethodGet, SuccessStatus: http.StatusOK}
-	response, err := b.ClientV1.ExecuteWithRetry(
```
Lagher0, line 500: So we getting rid of retries here? Why are we making changes to the retry logic in this PR?

internal/api/client.go
panigrahisubhrajit: 10 is somewhat arbitrary enough to absorb a short burst of 429s without burning the full 10-minute timeout. Open to a different value if you have a preference.

internal/resources/backup.go
panigrahisubhrajit: Removed , was explaining the change rather than the code. The function call makes the intent clear.

internal/resources/backup.go
panigrahisubhrajit: This isn't retry logic, it's the core bug fix. The old code only read page 1 of the backup list, so on a busy cluster the latest backup was on a later page and polling never detected our new backup. GetPaginated walks all pages. The maxRetryAttempts change in client.go is a separate secondary guard on 429 flaps, added while we were already touching the client.

internal/api/client.go
panigrahisubhrajit: Only affects calls that get 10 consecutive 429 or 504 responses. Any request that succeeds normally is untouched existing tests won't see any change.

internal/api/client.go
panigrahisubhrajit: For 429s we already use the server-provided Retry-After header, so backoff is server-controlled. For 504s it's a fixed 2s wait. Adding exponential for 504s is reasonable but out of scope here.

internal/api/client.go
panigrahisubhrajit: Execute predates this PR, we didn't add or change it here. You're right it looks unused will remove it in a follow-up.

internal/api/client.go
panigrahisubhrajit: Execute predates this PR, we didn't add or change it here. You're right it looks unused — will remove it in a follow-up.

internal/api/client.go
```diff
@@ -57,59 +57,10 @@ type EndpointCfg struct {
...
-) (response *Response, err error) {
-	var requestBody []byte
-	if payload != nil {
-		requestBody, err = json.Marshal(payload)
-		if err != nil {
-			return nil, fmt.Errorf("%s: %w", errors.ErrMarshallingPayload, err)
-		}
-	}
+// maxRetryAttempts caps 429/504 retries independently of the 10-minute wall-clock
+// timeout, so a rapid flap with small backoffs cannot exhaust the full window.
```
Lagher0: What does `rapid flap` mean?

internal/api/client.go
Lagher0: In what scenarios, do we expect bursts of 429 failures?

internal/api/client.go
Lagher0: When do we expect to encounter 504 errors for the tests? Won't we still have flaky tests, if we fail the test in 20 seconds of 504 error? How long would we expect service timeouts to last?

internal/resources/backup.go
Lagher0: What are 429 flaps?

internal/api/client.go
IsaacLambat: +1 this, saw a lot of mention of `flaps` and took a guess at what it means but really comments are meant to take the guess work out of what the code is doing, not the other way round.

panigrahisubhrajit: A rapid flap is when the API keeps returning 429 or 504 on back-to-back retries — e.g. hitting the rate limit on every attempt at 2s apart. Without a count cap, that would retry for the full 10 minutes.

internal/api/client.go
panigrahisubhrajit: On CI, multiple acceptance tests run in parallel and all hit the same Capella API. Under that load we occasionally get rate-limited in bursts. Rare in normal usage but can happen in heavy test runs.

internal/api/client.go
panigrahisubhrajit: 504s from Capella's gateway are usually transient, resolving in a few seconds. The 10-minute timeout is the real guard — maxAttempts just stops us hammering the API 300+ times with small backoffs before that window is up. If we're hitting 504s for 20 straight seconds that's a genuine service problem and failing early is the right call.

internal/api/client.go
```diff
@@ -234,9 +242,10 @@ func exec(
 	for {
 		select {
 		case <-ctx.Done():
-			return nil, fmt.Errorf("timed out executing request against api: %w", err)
+			return nil, fmt.Errorf("timed out executing request against api after %d attempts: %w", attempts, err)
```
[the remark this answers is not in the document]
panigrahisubhrajit: Already using ctx.Err() here, not err — so this can't return nil.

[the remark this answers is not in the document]
panigrahisubhrajit: Already using ctx.Err() here, not err — so this can't return nil.

internal/resources/backup.go
panigrahisubhrajit: when the API keeps returning 429 or 504 on back-to-back retries with small delays e.g. hitting the rate limit on every attempt at 2s apart. Without a count cap, that would keep retrying for the full 10 minutes.

internal/api/client.go
```diff
@@ -57,6 +57,12 @@ type EndpointCfg struct {
 // defaultWaitAttempt re-attempt http request after 2 seconds.
 const defaultWaitAttempt = time.Second * 2
 
+// maxRetryAttempts caps the number of times ExecuteWithRetry will resend a request
+// on 429/504. Combined with the 10-minute outer envelope below, this stops a
+// pathological flap from silently burning the test-binary timeout budget. Picked
+// so the worst case stays under ~2 min even when each attempt completes quickly.
```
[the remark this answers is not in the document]
panigrahisubhrajit: Fixed — renamed to maxAttempts and updated the error message to say 'exhausted %d attempts'.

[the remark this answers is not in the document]
panigrahisubhrajit: Fixed — updated the comment to accurately describe the behaviour.

internal/resources/backup.go
panigrahisubhrajit: Done — removed the prior-implementation section.

internal/resources/backup.go
```diff
@@ -402,53 +401,60 @@ func (a *Backup) validateCreateBackupRequest(plan providerschema.Backup) error {
...
+// waitForBackup polls the specific backup record (looked up by its server-assigned
+// id from the POST 202 response) until it reaches a final state (ready/failed) or
+// the timeout expires. Polling by id avoids the previous design's race against the
+// list endpoint's undefined default ordering and the per-bucket serialization
+// window: we know exactly which record we're waiting on.
+//
+// 30 min is a deliberately tighter cap than the prior 90-min ceiling — under the
+// 120m Go test budget, the suite can absorb several stuck backups instead of a
+// single one consuming the entire run. Users with very large buckets that need
+// longer should override this via a per-resource timeouts block (follow-up).
```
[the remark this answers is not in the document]
panigrahisubhrajit: Fixed — the comment about per-resource timeouts block has been removed.

internal/api/client.go
panigrahisubhrajit: 504s from Capella's gateway are usually transient — a slow operation that resolves in a few seconds. The 10-minute timeout is the real guard

internal/api/client.go
panigrahisubhrajit: On the CI tenant, multiple acceptance tests run in parallel and all hit the same Capella API at the same time. Under that load we do occasionally get rate-limited in bursts. It's rare in normal customer usage but can happen in heavy test runs.

internal/api/client.go
panigrahisubhrajit: A rapid flap is when the API keeps returning 429 or 504 on back-to-back retries with small delays

Lagher0: Where are you getting this terminology from? Could you please share the source?

internal/api/client.go
Lagher0: Since the value 10 is somewhat arbitrary by your own admission, it would help to anchor it to something observable — e.g. the max parallel test workers on the CI tenant, or the measured burst length from a past incident. Otherwise it's hard to know if 10 should be 5 or 20 in the future. What does a 'heavy test run' look like concretely?

internal/api/client.go
panigrahisubhrajit: I'll replace it with "burst of consecutive 429 responses" to keep the language plain.

internal/api/client.go
panigrahisubhrajit: Fair point,  there's no measurement behind it. The 10-minute context timeout is the real bound; maxAttempts is just an early exit that isn't grounded in anything concrete. Happy to remove it and let the timeout do the job on its own. That also simplifies the code.

panigrahisubhrajit: When multiple acceptance tests run in parallel against the same Capella tenant they share the same API rate limit — back-to-back requests from concurrent test workers can briefly hit the per-tenant 429 threshold. It's not common but it happens on busy CI runs.

internal/api/client.go
panigrahisubhrajit: 504s are rare — they happen when the control plane is briefly overloaded. You're right that the 10-attempt cap makes this worse: if 504s last longer than 20s the test fails unnecessarily. Removing maxAttempts and relying solely on the 10-minute timeout fixes that — 504s would keep retrying for the full window instead of bailing early. Will drop the cap.

Lagher0: So we don't want to implement the exponential backoffs and want to the fail fast for the 504s? Do we have any data on how long we hit 504s for? How do we ensure that we are waiting an appropriate amount of time before failing the test?

panigrahisubhrajit: After removing maxAttempts, 504s no longer fail fast. they retry with a 2s wait for the full 10 minutes (~300 attempts), same as 429s. No exponential backoff for 504s, and we don't have incident data on how long they last. The 2s fixed wait is a guess. If you'd prefer exponential backoff for 504s too, happy to add it, but given this is a bug fix PR for an unrelated issue, would you be ok leaving that as a follow-up? The current behaviour is strictly better than before (no hard 10-attempt cap).

internal/api/client.go
Lagher0: Why were you using a networking/ops term for retry logic?

internal/api/client.go
Lagher0: I'm happy to have it as a follow up. I'm trying to understand what is the ideal behaviour we want for the retry and what we are trying to achieve.

panigrahisubhrajit: The goal is to absorb transient errors so a terraform apply/read doesn't fail due to a momentary API blip. 

429: server tells us when to retry via Retry-After. 
504: control plane is briefly overloaded , fixed 2s wait, keep retrying until the 10-minute timeout
The 10-minute timeout is the outer bound. If things haven't recovered in 10 minutes it's not transient and we should fail. Exponential backoff for 504s would be a nice-to-have but the current fixed 2s is safe enough for now.

Lagher0 (commented):
Overall looks the code good to me. What are the other failing acceptance tests?

internal/api/client.go
panigrahisubhrajit: The code is about HTTP retry logic and the comment should use HTTP terminology. It's gone now anyway since we removed maxAttempts entirely.

panigrahisubhrajit:
> Overall looks the code good to me. What are the other failing acceptance tests?

From what I can tell, couchbase-capella_backup is the only resource where getLatestBackup was doing a page-1 only lookup. Other resources that poll for state (cluster, app service, etc.) poll by a known ID directly, so they don't have the same issue. The only test failing in the current CI run is TestAccAppServiceLogStreaming, which is unrelated to this PR and was already failing before these changes.

IsaacLambat (approved):
Failing CI does look unrelated, do you know if there’s a ticket for fixing the broken test?

Edit: Talked to @panigrahisubhrajit and it seems the test is fixed in https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/604

PR #608, opened 2026-05-21
[AV-129986] Add updateDeletionProtection method and request payload for cluster deletion protection

stanleefdz:
## Jira

* AV-129986

## Description

Add updateDeletionProtection method and request payload for cluster deletion protection

## Type of Change

- [x] Manually tested
- [ ] Unit tested
- [ ] Acceptance tested

### Testing

<details open>
  <summary>Testing</summary>
  
</details>

## Further comments

### DIFF
--- a/internal/resources/cluster.go
+++ b/internal/resources/cluster.go
@@ -368,7 +368,7 @@ func (c *Cluster) Update(ctx context.Context, req resource.UpdateRequest, resp *
 		return
 	}
 
-	if !plan.DeletionProtection.Equal(state.DeletionProtection) {
+	if !plan.DeletionProtection.IsUnknown() && !plan.DeletionProtection.IsNull() && !plan.DeletionProtection.Equal(state.DeletionProtection) {
 		if err := c.updateDeletionProtection(ctx, organizationId, projectId, clusterId, plan.DeletionProtection.ValueBool()); err != nil {
 			resp.Diagnostics.AddError(
 				"Error updating cluster deletion protection",
--- a/internal/resources/cluster.go
+++ b/internal/resources/cluster.go
@@ -109,6 +109,10 @@ func (c *Cluster) Create(ctx context.Context, req resource.CreateRequest, resp *
 		clusterRequest.EnablePrivateDNSResolution = plan.EnablePrivateDNSResolution.ValueBoolPointer()
 	}
 
+	if !plan.DeletionProtection.IsNull() && !plan.DeletionProtection.IsUnknown() {
+		clusterRequest.DeletionProtection = plan.DeletionProtection.ValueBoolPointer()
+	}
+
 	if plan.Zones != nil && (plan.CloudProvider.Type.ValueString() != string(clusterapi.Aws) || plan.Availability.Type.ValueString() != "single") {
 		resp.Diagnostics.AddError(
 			"Error creating cluster",
@@ -897,6 +901,10 @@ func initializePendingClusterWithPlanAndId(plan providerschema.Cluster, id strin
 
 	if plan.EnablePrivateDNSResolution.IsNull() || plan.EnablePrivateDNSResolution.IsUnknown() {
 		plan.EnablePrivateDNSResolution = types.BoolNull()
+	}
+
+	if plan.DeletionProtection.IsNull() || plan.DeletionProtection.IsUnknown() {
+		plan.DeletionProtection = types.BoolNull()
 	}
 
 	if plan.CouchbaseServer.IsNull() || plan.CouchbaseServer.IsUnknown() {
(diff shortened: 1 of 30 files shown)


internal/resources/cluster.go
```diff
@@ -671,6 +671,20 @@ func (c *Cluster) checkClusterStatus(ctx context.Context, organizationId, projec
...
+	_, err := c.ClientV1.ExecuteWithRetry(
+		ctx,
+		cfg,
+		clusterapi.UpdateDeletionProtectionRequest{DeletionProtection: enabled},
+		c.Token,
+		nil,
+	)
+	return err
+}
+
```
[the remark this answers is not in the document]
stanleefdz: fixed it in [abe5f5b](https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/608/commits/abe5f5bf784e724ba783ceb8656e3fc0f932f50d)

internal/resources/cluster.go
```diff
@@ -368,6 +368,16 @@ func (c *Cluster) Update(ctx context.Context, req resource.UpdateRequest, resp *
 		return
 	}
 
+	if !plan.DeletionProtection.Equal(state.DeletionProtection) {
```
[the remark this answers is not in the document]
stanleefdz: fixed in [ffb0057](https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/608/commits/ffb0057bacc982220269684a7d1e6ac4603d42fc)

internal/resources/cluster_deletion_protection.go
```diff
@@ -0,0 +1,201 @@
...
+
+func (r *ClusterDeletionProtection) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
+	var plan providerschema.ClusterDeletionProtection
+	diags := req.Plan.Get(ctx, &plan)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if err := r.putDeletionProtection(ctx, plan.OrganizationId.ValueString(), plan.ProjectId.ValueString(), plan.ClusterId.ValueString(), plan.DeletionProtection.ValueBool()); err != nil {
```
IsaacLambat, line 67: Will this cause a problem is deletion protection is already set on the cluster, and then a user tries to create this resource with it enabled?

Would be good maybe to have a bit more manual testing regarding cases like these

internal/resources/cluster_schema.go
```diff
@@ -20,7 +20,7 @@ func ClusterSchema() schema.Schema {
 	capellaschema.AddAttr(attrs, "description", clusterBuilder, stringAttribute([]string{optional, computed}))
 	capellaschema.AddAttr(attrs, "zones", clusterBuilder, stringSetAttribute(optional, requiresReplace))
 	capellaschema.AddAttr(attrs, "enable_private_dns_resolution", clusterBuilder, boolDefaultAttribute(false, optional, computed, requiresReplace))
-	capellaschema.AddAttr(attrs, "deletion_protection", clusterBuilder, boolAttribute(optional, computed))
+	capellaschema.AddAttr(attrs, "deletion_protection", clusterBuilder, boolAttribute(optional, computed, useStateForUnknown))
```
IsaacLambat: Why is `useStateForUnknown` needed here? This value can be changed underneath TF so I don't think this is appropriate here. This tag should only be used when the value will never change once the resource has been created (such as an ID)

Edit: Just seen that my comment conflicts with https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/608#discussion_r3285316460, but still curious to get your thoughts here

internal/resources/cluster_deletion_protection.go
stanleefdz: PUT operation works in all the scenarios to update the existing field value to the latest one, there is no restriction

internal/resources/cluster_schema.go
stanleefdz: I added it coz of the guard rail pointed out by droid

IsaacLambat (commented):
Just one more question. I notice that the cluster schema has a `deletion_protection` field. Will this cause a conflict with this resource if both are set to different values? I.E if a user sets `deletion_protection = true` on the cluster resource and false on this resource, what happens on apply?

stanleefdz:
> Just one more question. I notice that the cluster schema has a `deletion_protection` field. Will this cause a conflict with this resource if both are set to different values? I.E if a user sets `deletion_protection = true` on the cluster resource and false on this resource, what happens on apply?

fixed it in [2c74289](https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/608/commits/2c7428973cec111370567403ef021eca58f7110c)
The couchbase-capella_cluster_deletion_protection resource is now the sole owner of the deletion protection write path, eliminating conflicts when both resources target the same API endpoint with different values. The cluster resource still reads the value (computed) and blocks deletion when it's enabled.

IsaacLambat (approved):

PR #630, opened 2026-06-03
[AV-132227] Add config validation for cluster on/off schedule days

PaulomeeCb:
## Jira

* AV-132227

## Description

Fixes [AV-132227](https://jira.issues.couchbase.com/browse/AV-132227): `couchbase-capella_cluster_onoff_schedule` accepted a schedule where every day is set to `off`, even though the Capella API forbids clusters being off for the entire day for every day of the week. `terraform validate` passed and the invalid config was only rejected by the API at apply time.

## Root Cause

The resource schema only validates each `day` value individually (`stringvalidator.OneOf`). There was no cross-item validation on the full `days` list — count, ordering, and the all-days-off constraint were left entirely to the API.

## Changes

- **`internal/resources/cluster_onoff_schedule.go`**: Implement `resource.ResourceWithValidateConfig` so `terraform validate` fails locally when:
  - the `days` list does not contain exactly 7 entries, one per day of the week
  - the days are not in the API-required Monday-to-Sunday sequence
  - every day's state is `off`

  Unknown/computed day values (e.g. built from other resources' outputs) defer validation to apply time instead of erroring spuriously.

- **`acceptance_tests/cluster_onoff_schedule_acceptance_test.go`**: Add acceptance tests:
  - `TestAccClusterOnOffScheduleAllDaysOff` — all-days-off schedule is rejected with a local validation error
  - `TestAccClusterOnOffScheduleOutOfOrderDays` — out-of-sequence days are rejected

  The existing `TestAccClusterOnOffScheduleResourceInvalidDays` regex still matches the new local 7-day error message.

## Type of Change

- [x] Manually tested
- [ ] Unit tested
- [x] Acceptance tested

### Testing

<details open>
  <summary>Testing</summary>

Ran the new acceptance tests against a sandbox organization with an existing project and cluster (`TF_VAR_project_id` / `TF_VAR_cluster_id` set). Both tests fail at plan time inside local config validation, so no schedule request ever reaches the Capella API:

```
$ TF_ACC=1 go test ./acceptance_tests/ -run 'TestAccClusterOnOffScheduleAllDaysOff|TestAccClusterOnOffScheduleOutOfOrderDays' -v -timeout 30m

=== RUN   TestAccClusterOnOffScheduleAllDaysOff
=== PAUSE TestAccClusterOnOffScheduleAllDaysOff
=== RUN   TestAccClusterOnOffScheduleOutOfOrderDays
=== PAUSE TestAccClusterOnOffScheduleOutOfOrderDays
=== CONT  TestAccClusterOnOffScheduleAllDaysOff
=== CONT  TestAccClusterOnOffScheduleOutOfOrderDays
--- PASS: TestAccClusterOnOffScheduleOutOfOrderDays (1.97s)
--- PASS: TestAccClusterOnOffScheduleAllDaysOff (1.98s)
PASS
ok  	github.com/couchbasecloud/terraform-provider-couchbase-capella/acceptance_tests	791.156s
```

The all-days-off config now fails `terraform plan`/`validate` with:

```
Error: Invalid Cluster On/Off Schedule

Clusters cannot be scheduled to be off for the entire day for every day of
the week. At least one day must have state "on" or "custom".
```

Also verified: `goimports`/`go vet` clean and the provider binary builds with version ldflags.

</details>

## Further comments

[AV-132227]: https://couchbasecloud.atlassian.net/browse/AV-132227?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/acceptance_tests/cluster_onoff_schedule_acceptance_test.go
+++ b/acceptance_tests/cluster_onoff_schedule_acceptance_test.go
@@ -213,10 +213,9 @@ resource "couchbase-capella_cluster_onoff_schedule" "%[2]s" {
 	})
 }
 
-// TestAccClusterOnOffScheduleOutOfOrderDays_AV_132227 verifies that a schedule
-// whose days are not in Monday-to-Sunday order is rejected by local config
-// validation (https://jira.issues.couchbase.com/browse/AV-132227).
-func TestAccClusterOnOffScheduleOutOfOrderDays_AV_132227(t *testing.T) {
+// TestAccClusterOnOffScheduleOutOfOrderDays verifies that a schedule
+// whose days are not in Monday-to-Sunday order is rejected by local config validation.
+func TestAccClusterOnOffScheduleOutOfOrderDays(t *testing.T) {
 	resourceName := randomStringWithPrefix("tf_acc_cluster_onoff_schedule_out_of_order_")
 
 	resource.ParallelTest(t, resource.TestCase{
--- a/acceptance_tests/cluster_onoff_schedule_acceptance_test.go
+++ b/acceptance_tests/cluster_onoff_schedule_acceptance_test.go
@@ -206,7 +206,9 @@ resource "couchbase-capella_cluster_onoff_schedule" "%[2]s" {
   ]
 }
 `, globalProviderBlock, resourceName, globalOrgId, globalProjectId, globalClusterId),
-				ExpectError: regexp.MustCompile(`off\s+for\s+the\s+entire\s+day\s+for\s+every\s+day\s+of\s+the\s+week`),
+				// Subset from the start of the message so Terraform's diagnostic
+				// word-wrapping cannot split the match.
+				ExpectError: regexp.MustCompile(`Clusters cannot be scheduled`),
 			},
 		},
 	})
@@ -240,7 +242,9 @@ resource "couchbase-capella_cluster_onoff_schedule" "%[2]s" {
   ]
 }
 `, globalProviderBlock, resourceName, globalOrgId, globalProjectId, globalClusterId),
-				ExpectError: regexp.MustCompile(`sequence\s+starting\s+from\s+Monday\s+and\s+ending\s+with\s+Sunday`),
+				// Subset from the start of the message so Terraform's diagnostic
+				// word-wrapping cannot split the match.
+				ExpectError: regexp.MustCompile(`must be in sequence`),
 			},
 		},
 	})
--- a/acceptance_tests/cluster_onoff_schedule_acceptance_test.go
+++ b/acceptance_tests/cluster_onoff_schedule_acceptance_test.go
@@ -178,6 +178,76 @@ resource "couchbase-capella_cluster_onoff_schedule" "%[2]s" {
 	})
 }
 
+// TestAccClusterOnOffSchedule_AV_132227 verifies that a schedule with every day
+// set to "off" is rejected by local config validation instead of being sent to
+// the API (https://jira.issues.couchbase.com/browse/AV-132227).
+func TestAccClusterOnOffSchedule_AV_132227(t *testing.T) {
+	resourceName := randomStringWithPrefix("tf_acc_cluster_onoff_schedule_all_off_")
+
+	resource.ParallelTest(t, resource.TestCase{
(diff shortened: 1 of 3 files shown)


acceptance_tests/cluster_onoff_schedule_acceptance_test.go
```diff
@@ -178,6 +178,74 @@ resource "couchbase-capella_cluster_onoff_schedule" "%[2]s" {
...
+    { day = "tuesday",   state = "off" },
+    { day = "wednesday", state = "off" },
+    { day = "thursday",  state = "off" },
+    { day = "friday",    state = "off" },
+    { day = "saturday",  state = "off" },
+    { day = "sunday",    state = "off" },
+  ]
+}
+`, globalProviderBlock, resourceName, globalOrgId, globalProjectId, globalClusterId),
+				ExpectError: regexp.MustCompile(`off\s+for\s+the\s+entire\s+day\s+for\s+every\s+day\s+of\s+the\s+week`),
```
IsaacLambat: nit: could this be made more readable by putting just a subset of the error message without `\s` like we do with other expectErrors? Same with L243

internal/resources/cluster_onoff_schedule.go
```diff
@@ -21,11 +21,16 @@ import (
...
-	_ resource.ResourceWithImportState = &ClusterOnOffSchedule{}
+	_ resource.Resource                   = &ClusterOnOffSchedule{}
+	_ resource.ResourceWithConfigure      = &ClusterOnOffSchedule{}
+	_ resource.ResourceWithImportState    = &ClusterOnOffSchedule{}
+	_ resource.ResourceWithValidateConfig = &ClusterOnOffSchedule{}
 )
 
+// weekdays is the order the V4 API requires for the on/off schedule days list:
+// exactly one entry per day of the week, starting from Monday and ending with Sunday.
+var weekdays = [...]string{"monday", "tuesday", "wednesday", "thursday", "friday", "saturday", "sunday"}
```
IsaacLambat: nit: we could define 7 in this so we have a bit more compiler time guarding incase someone accidently adds or removes a value from this

internal/resources/cluster_onoff_schedule.go
```diff
@@ -55,6 +60,66 @@ func (c *ClusterOnOffSchedule) Schema(_ context.Context, _ resource.SchemaReques
 	resp.Schema = OnOffScheduleSchema()
 }
 
+// ValidateConfig enforces the cross-item constraints on the days list that the
+// V4 API documents but the attribute-level validators cannot express: the
+// schedule must contain exactly one entry per day of the week in Monday-to-Sunday
+// order, and the cluster cannot be scheduled to be off for every day of the week.
+func (c *ClusterOnOffSchedule) ValidateConfig(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
```
IsaacLambat, line 68: Assuming enum validation has not been added to the schema due to the plan is to automatically do that using the api specs? ie. `stringvalidator.OneOf("on", "off", "custom")` to `internal/resources/cluster_onoff_schedule_schema.go L26`?

IsaacLambat (dismissed):

acceptance_tests/cluster_onoff_schedule_acceptance_test.go
PaulomeeCb: Done — replaced with short subsets (Clusters cannot be scheduled / must be in sequence) taken from the start of each message so diagnostic word-wrapping can't split the match.

internal/resources/cluster_onoff_schedule.go
PaulomeeCb: Done — changed [...]string to [7]string so adding/removing an entry is a compile error.

internal/resources/cluster_onoff_schedule.go
PaulomeeCb: Good catch — added stringvalidator.OneOf("on", "off", "custom") to the state attribute in the schema.

IsaacLambat (approved):

PR #638, opened 2026-06-11
[AV-134037] Auto-attach enum validators for -to-named-enum fields

PaulomeeCb:
<!-- REMINDER: All testing and verification for your change should be completed within localdev or a sandbox environment.
ONLY MERGE INTO MAIN IF CHANGES ARE TESTED, VERIFIED AND QUALITY CHECKED THOROUGHLY

Merging to main == merging to PRODUCTION.
-->

## Jira

* AV-134037
* AV-132232

## Description

 ### Problem                                                                                                                                                                                    
                                                                                                                                                                                                 
  The enum generator silently dropped enum metadata for any schema property whose OpenAPI type is a **`$ref` to a named scalar-enum schema**, so `SchemaBuilder.AddAttr` attached no `OneOf` validator for those fields. Invalid values passed `terraform validate` and were only rejected later by the API. This is the root cause behind AV-132232 (cluster on/off schedule accepted an empty `timezone`), and it affected other fields too.                                                                                                                                           
                                                                                                                                                                                                 
  ### Root cause                                                                                                                                                                                 
                                                                                                                                                                                                 
  1. `discover.go` skips property `$ref`s (`propRef.Ref != "" -> continue`), assuming the target schema is captured independently as a top-level component.                                      
  2. In the spec, `ClusterOnOffSchedule.timezone` is `{"$ref": "#/components/schemas/onOffTimezone"}` (a shared `string` enum), so the property is skipped.                                      
  3. `onOffTimezone` is visited top-level, but as a bare enum it produces a site with an **empty field path**, which `buildEnumTable` discards.                                                  
                                                                                                                                                                                                 
The enum reached neither the property nor the table. (`ColumnarAnalyticsOnOffSchedule.timezone` only worked because the spec defines it **inline**.)    

### Change                                                                                                                                                                                     
                                                                                                                                                                                                 
  - `discover.go`: capture the enum on a property/array item when its `$ref` target is a scalar enum, via a new `refEnum` helper. Object `$ref`s are still walked top-level (unchanged).         
  - Regenerated `internal/generated/enums/enums.gen.go` — **24 previously missing entries** are now populated, so the builder auto-attaches `OneOf` validators (no hand-coded lists).            
  - Added `TestAccClusterOnOffScheduleEmptyTimezone`; tightened the invalid-timezone test regex to the `OneOf` message.  

<!-- What does this change do? Why is it needed? -->

## Type of Change

## Manual Testing Approach

- [x] Manually tested
- [ ] Unit tested
- [x] Acceptance tested

### Testing

<details open>
  <summary>Testing</summary>

**Generator / unit:** `go test ./internal/generated/enums/...` passes; versioned build and `go vet` clean. The regeneration populated 24 previously-missing enum entries.

**`terraform validate`** (dev-override build) across every newly-enforced writable field:

- Invalid values rejected locally with the `OneOf` diagnostic:
  - `timezone = ""` (cluster on/off schedule) → `Attribute timezone value must be one of: ["Pacific/Midway" ...]`
  - `bucket_conflict_resolution = "bogus"` → `must be one of: ["seqno" "lww"]`
  - `configuration_type = "tripleNode"` → `must be one of: ["singleNode" "multiNode"]`
  - `support.timezone = "ZZ"` → `must be one of: ["ET" "GMT" "IST" "PT"]`
- Valid values pass — `US/Pacific`, `seqno`, `multiNode`, `PT` → `Success! The configuration is valid`.

**Acceptance tests** — run against a live Capella sandbox (`TF_ACC=1`):

```
--- PASS: TestAccClusterOnOffScheduleResourceInvalidTimezone (2.16s)
--- PASS: TestAccClusterOnOffScheduleEmptyTimezone (2.16s)
--- PASS: TestAccBucketResourceRequiredOnly (5.12s)
PASS
ok  	github.com/couchbasecloud/terraform-provider-couchbase-capella/acceptance_tests	854.280s
```

`TestAccClusterOnOffScheduleEmptyTimezone` confirms the timezone enum is now validated automatically (no hand-coded list — the validator is attached by the generator-populated table). `TestAccBucketResourceRequiredOnly` confirms the newly-enforced `bucket_conflict_resolution` default does not break bucket creation.

</details>

## Further comments

### DIFF
--- a/internal/generated/enums/enums.gen.go
+++ b/internal/generated/enums/enums.gen.go
@@ -43,6 +43,9 @@ var enumTable = map[string]map[string]EnumDef{
 	"CloudProvider": {
 		"type": {Type: "string", Values: []string{"aws", "gcp", "azure"}},
 	},
+	"ClusterOnOffSchedule": {
+		"timezone": {Type: "string", Values: []string{"Pacific/Midway", "US/Hawaii", "US/Alaska", "US/Pacific", "US/Mountain", "US/Central", "US/Eastern", "America/Puerto_Rico", "Canada/Newfoundland", "America/Argentina/Buenos_Aires", "Atlantic/Cape_Verde", "Europe/London", "Europe/Amsterdam", "Europe/Athens", "Africa/Nairobi", "Asia/Tehran", "Indian/Mauritius", "Asia/Karachi", "Asia/Calcutta", "Asia/Dhaka", "Asia/Bangkok", "Asia/Hong_Kong", "Asia/Tokyo", "Australia/North", "Australia/Sydney", "Pacific/Ponape", "Antarctica/South_Pole"}},
+	},
 	"ColumnarAnalyticsOnOffSchedule": {
 		"timezone": {Type: "string", Values: []string{"Pacific/Midway", "US/Hawaii", "US/Alaska", "US/Pacific", "US/Mountain", "US/Central", "US/Eastern", "America/Puerto_Rico", "Canada/Newfoundland", "America/Argentina/Buenos_Aires", "Atlantic/Cape_Verde", "Europe/London", "Europe/Amsterdam", "Europe/Athens", "Africa/Nairobi", "Asia/Tehran", "Indian/Mauritius", "Asia/Karachi", "Asia/Calcutta", "Asia/Dhaka", "Asia/Bangkok", "Asia/Hong_Kong", "Asia/Tokyo", "Australia/North", "Australia/Sydney", "Pacific/Ponape", "Antarctica/South_Pole"}},
 	},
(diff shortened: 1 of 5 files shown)


internal/generated/enums/enums.gen.go
```diff
@@ -43,6 +43,9 @@ var enumTable = map[string]map[string]EnumDef{
 	"CloudProvider": {
```
matty271828, line 43: Can we please commit the generated result under a separate ticket.

matty271828 (commented):
Logic looks fine and seems to have pulled in the additional enums okay. Can you please update the unit tests in `discover_test.go` and `generator_test.go`  to cover this behavior?

internal/generated/enums/enums.gen.go
PaulomeeCb: Hey @matty271828 thanks for the review! I have added the unit tests. I can move it to a separate ticket - just want to understand why is it needed?  Is it because generated files should be updated in their own PR?                                                                                

Note: if I take `enums.gen.go` out of this PR, the `discover.go` change won't actually do anything on its own (nothing changes for users until the file is regenerated), and the empty-timezone test would move to that other ticket. That's why I preferred keeping enums.gen.go here. Does that make sense? Lmk

matty271828: Just slightly better housekeeping imo. If we ever wanted to revert the commit it would be cleaner. Not a big deal though happy to approve.

matty271828 (commented):
LGTM. Note there is acceptance test failure though please raise a ticket to track the issue.

PaulomeeCb:
> LGTM. Note there is acceptance test failure though please raise a ticket to track the issue.

Done. https://jira.issues.couchbase.com/browse/AV-134307

nidhi07kumar (approved):
LGTM

PR #640, opened 2026-06-12
[AV-131313] Add load balancer CIDR support to app service

JRascagneres:
## Jira

* AV-131313

## Description

Implements terraform side of the user defined CIDR option. New `loadBalancerCidr` option for App Services Azure clusters allowing the customer to define the cidr for the load balancer to be deployed with.

## Type of Change

- [X] New feature (non-breaking change which adds functionality).

- [X] Manually tested
- [ ] Unit tested
- [X] Acceptance tested

### Testing

<details open>
<summary>Testing</summary>
Written two acceptance tests that test it pretty end to end. Success proof from GoLand run. Will fail CI. Needs the FF enabled:
image

Also did a deployment myself to validate:
image
image

</details>

## Further comments

### DIFF
--- a/acceptance_tests/app_service_load_balancer_cidr_acceptance_test.go
+++ b/acceptance_tests/app_service_load_balancer_cidr_acceptance_test.go
@@ -1,0 +1,283 @@
+package acceptance_tests
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"regexp"
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+
+	"github.com/couchbasecloud/terraform-provider-couchbase-capella/internal/api"
+	"github.com/couchbasecloud/terraform-provider-couchbase-capella/internal/api/appservice"
+)
+
+// TestAccAppServiceLoadBalancerCIDR tests the load_balancer_cidr attribute on the
+// app service resource and data source against a dedicated Azure cluster.
+func TestAccAppServiceLoadBalancerCIDR(t *testing.T) {
+	// load_balancer_cidr is Azure only and must be a valid IPv4 /24 network address.
+	// The global cluster is AWS, so this test stands up its own Azure cluster. The
+	// 192.168.x.0/24 ranges cannot overlap the random 10.x.y.0/23 cluster CIDR, and
+	// the control plane adds the /24 as an extra VNet address space, so the app
+	// service deploys cleanly.
+	const (
+		loadBalancerCIDR        = "192.168.0.0/24"
+		loadBalancerCIDRChanged = "192.168.1.0/24"
+	)
+
+	resourceName := randomStringWithPrefix("tf_acc_app_svc_lb_")
+	resourceReference := "couchbase-capella_app_service." + resourceName
+	clusterName := randomStringWithPrefix("tf_acc_cluster_azure_")
+	dataSourceName := randomStringWithPrefix("tf_acc_app_svcs_lb_ds_")
+	dataSourceReference := "data.couchbase-capella_app_services." + dataSourceName
+	appServiceName := randomStringWithPrefix("tf_acc_app_svc_lb_")
+	clusterCIDR := generateRandomCIDR()
+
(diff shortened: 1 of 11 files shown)


internal/resources/appservice.go
```diff
@@ -204,6 +215,12 @@ func (a *AppService) Read(ctx context.Context, req resource.ReadRequest, resp *r
 		refreshedState.IfMatch = state.IfMatch
 	}
 
+	// Preserve the configured load_balancer_cidr; the API omits it unless Azure + defined-CIDR is
+	// enabled. On import there is no prior value, so the refreshed (API) value is kept.
+	if !state.LoadBalancerCidr.IsNull() && !state.LoadBalancerCidr.IsUnknown() {
+		refreshedState.LoadBalancerCidr = state.LoadBalancerCidr
+	}
```
IsaacLambat: Maybe a todo ticket regarding removing this in the future once all app services have upgraded after the FF is flipped?

internal/resources/appservice.go
```diff
@@ -311,6 +328,12 @@ func (a *AppService) Update(ctx context.Context, req resource.UpdateRequest, res
 		currentState.IfMatch = plan.IfMatch
 	}
 
+	// load_balancer_cidr is create-only (requires replace), so it never changes here. Keep the
+	// configured value since the API omits it unless Azure + defined-CIDR is enabled.
+	if !plan.LoadBalancerCidr.IsNull() && !plan.LoadBalancerCidr.IsUnknown() {
+		currentState.LoadBalancerCidr = plan.LoadBalancerCidr
+	}
```
IsaacLambat: Not seeing what this is meant to do. If the plan does have a CIDR defined, then it should already be in the currentState that is returned (as the API will not omit it due to using Azure + defined-CIDR being enabled).

If the user changes/adds a LB CIDR then this function should never be hit, and instead delete then create will be hit due to the `requires replace`.

Let me know what I am missing here :D

JRascagneres: Yeah true. We can't hit this scenario if once an app service has been created with a user-defined cidr a GET always returns the field... Will remove. Can probably remove others too. I was trying to handle a scenario where the FF was turned off again but thats not really doable.

acceptance_tests/app_service_load_balancer_cidr_acceptance_test.go
```diff
@@ -0,0 +1,283 @@
...
+		}
+
+		return fmt.Errorf("app service %q not found in %s.data across %d entries", appServiceID, dataSourceReference, count)
+	}
+}
+
+// testAccCheckAppServiceHealthy blocks until the app service reaches the healthy
+// state. The provider's create can return before the deploy finishes, so this
+// keeps the later replace and teardown from deleting a still-deploying app service.
+func testAccCheckAppServiceHealthy(resourceReference string) resource.TestCheckFunc {
```
IsaacLambat: I thought the App Service resource creation blocks until creation anyway? Making this not needed?

JRascagneres: Okay looks good removed

image

IsaacLambat (dismissed):

IsaacLambat (approved):

IsaacLambat (dismissed):

docs/resources/app_service.md
```diff
@@ -51,6 +51,7 @@ resource "couchbase-capella_app_service" "new_app_service" {
 
 - `description` (String) - A short description of the App Service.
 - `if_match` (String) A precondition header that specifies the entity tag of a resource.
+- `load_balancer_cidr` (String)
 - `nodes` (Number) - Number of nodes configured for the App Service. Number of nodes configured for the App Service. The number of nodes can range from 2 to 12.
```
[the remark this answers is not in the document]
IsaacLambat: Ignore, as this is generated documentation that will be overwritten

docs/data-sources/app_services.md
```diff
@@ -45,6 +45,7 @@ Read-Only:
  - **Constraints**: Maximum length: 1024 characters
 - `id` (String) - The ID of the App Service created.
  - **Format**: UUID (GUID4)
+- `load_balancer_cidr` (String)
 - `name` (String) - Name of the App Service (up to 256 characters).
  - **Constraints**: Maximum length: 256 characters
```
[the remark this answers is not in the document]
IsaacLambat: Ignore, as this is generated documentation that will be overwritten

IsaacLambat (approved):

PR #660, opened 2026-06-22
[AV-132168] Reject unsupported network peer provider_type

PaulomeeCb:
<!-- REMINDER: All testing and verification for your change should be completed within localdev or a sandbox environment.
ONLY MERGE INTO MAIN IF CHANGES ARE TESTED, VERIFIED AND QUALITY CHECKED THOROUGHLY

Merging to main == merging to PRODUCTION.
-->

## Jira

* AV-132168

## Description

 ### Problem                                                                                                                                                                                    
  `couchbase-capella_network_peer.provider_type` accepted unsupported values (e.g. `provider_type = "ibm"`) during `terraform validate`. The error only surfaced later during create, when the API rejected it.                                                                                                                                                                               
                                                                                                                                                                                                 
  ### Root cause                                                                                                                                                                                 
  `provider_type` was defined as a plain required string in `network_peer_schema.go`. Unlike other fields, the auto-attach mechanism can't help here: the generated enum table has no `providerType` entry because the OpenAPI spec (`CreateNetworkPeeringRequest.providerType`) doesn't expose it as an enum, so no `OneOf` validator gets attached.                                
                                                                                                                                                                                                 
  ### Fix                                                                                                                                                                                        
  Add an explicit `stringvalidator.OneOf("aws", "gcp", "azure")` to `provider_type`, matching the supported values documented in the schema model (`internal/schema/network_peer.go`) and the resource examples. Invalid values are now rejected at plan time, before any API call.  

<!-- What does this change do? Why is it needed? -->

## Type of Change

## Manual Testing Approach

- [ ] Manually tested
- [ ] Unit tested
- [X] Acceptance tested

### Testing

<details open>
  <summary>Testing</summary>
Acceptance tested against production Capella. TestMain provisioned a cluster, bucket, app service, and app endpoint, the plan-time validation test ran, and all fixtures were torn down (verified 0 leftover clusters).

```
=== RUN   TestAccNetworkPeerInvalidProviderType
--- PASS: TestAccNetworkPeerInvalidProviderType (1.54s)
PASS
ok  github.com/couchbasecloud/terraform-provider-couchbase-capella/acceptance_tests  851.216s
```

Asserts that `provider_type = "ibm"` is rejected at plan time (before any API call):
`Attribute provider_type value must be one of: ["aws" "gcp" "azure"], got: "ibm"`
</details>

## Further comments

### DIFF
--- a/acceptance_tests/network_peer_acceptance_test.go
+++ b/acceptance_tests/network_peer_acceptance_test.go
@@ -1,0 +1,48 @@
+package acceptance_tests
+
+import (
+	"fmt"
+	"regexp"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+)
+
+// TestAccNetworkPeerInvalidProviderType verifies that the network peer resource rejects an
+// unsupported provider_type at plan time. The OneOf("aws", "gcp", "azure") validator fires
+// before any API call, so dummy org/project/cluster IDs are sufficient.
+func TestAccNetworkPeerInvalidProviderType(t *testing.T) {
+	resourceName := randomStringWithPrefix("tf_acc_network_peer_invalid_")
+
+	resource.ParallelTest(t, resource.TestCase{
+		ProtoV6ProviderFactories: globalProtoV6ProviderFactory,
+		Steps: []resource.TestStep{
+			{
+				Config:      testAccNetworkPeerInvalidProviderTypeConfig(resourceName),
+				ExpectError: regexp.MustCompile(`(?s)provider_type.*value must be one of.*aws.*gcp.*azure`),
+			},
+		},
+	})
+}
+
+func testAccNetworkPeerInvalidProviderTypeConfig(resourceName string) string {
+	return fmt.Sprintf(`
+%[1]s
+
+resource "couchbase-capella_network_peer" "%[2]s" {
+  organization_id = "00000000-0000-0000-0000-000000000000"
+  project_id      = "11111111-1111-1111-1111-111111111111"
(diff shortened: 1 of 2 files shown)


IsaacLambat (approved):
```
Unlike other fields, the auto-attach mechanism can't help here: the generated enum table has no providerType entry because the OpenAPI spec (CreateNetworkPeeringRequest.providerType) doesn't expose it as an enum, so no OneOf validator gets attached
```

Doesn't this mean that the specs should be changed instead of the TF code working around this?

PaulomeeCb:
> ```
> Unlike other fields, the auto-attach mechanism can't help here: the generated enum table has no providerType entry because the OpenAPI spec (CreateNetworkPeeringRequest.providerType) doesn't expose it as an enum, so no OneOf validator gets attached
> ```
> 
> Doesn't this mean that the specs should be changed instead of the TF code working around this?

Agreed - fixing the spec is the cleaner fix, but it's upstream/generated and out of scope here. The explicit OneOf is the builder's intended call-site override (auto-attach is skipped when a validator is already present), so it closes the plan-time gap now. Filed [AV-135386](https://jira.issues.couchbase.com/browse/AV-135386) to add the enum to the spec; this validator can be removed once that lands.

[AV-135386]: https://couchbasecloud.atlassian.net/browse/AV-135386?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #664, opened 2026-06-24
[AV-135189] Update function on keyspace change

l0n3star:
## Jira

* AV-135189

## Description

this PR fixes the following:

a.  apply update when keyspace changes
b.  fix check to determine when eventing function changes

## Type of Change

- [X] Manually tested
- [ ] Unit tested
- [ ] Acceptance tested

### Testing

<details open>
  <summary>Testing</summary>
  
</details>

## Further comments

### DIFF
--- a/internal/resources/eventing_function.go
+++ b/internal/resources/eventing_function.go
@@ -375,10 +375,11 @@ func (e *EventingFunction) Update(ctx context.Context, req resource.UpdateReques
 		name           = IDs[providerschema.FunctionName]
 	)
 
+	eventingFunctionHasChanged := eventingFunctionChanged(&plan, &state, plannedSettings, stateSettings)
+
 	// eventing function can only be changed while the function is undeployed or paused.
 	// in other states block any changes up front to prevent "inconsistent result after apply" errors.
-	if (plan.State.ValueString() != eventingStateUndeployed && plan.State.ValueString() != eventingStatePaused) &&
-		eventingFunctionChanged(&plan, &state, plannedSettings, stateSettings) {
+	if (plan.State.ValueString() != eventingStateUndeployed && plan.State.ValueString() != eventingStatePaused) && eventingFunctionHasChanged {
 		resp.Diagnostics.AddError(
 			"Cannot change eventing function while deployed",
 			"Eventing function "+name+" must be in an undeployed or paused state in order to be changed."+
@@ -406,7 +407,9 @@ func (e *EventingFunction) Update(ctx context.Context, req resource.UpdateReques
 	}
 
 	// if only eventing function state was changed skip update step.
-	if plan.State.ValueString() == eventingStateDeployed || plan.State.ValueString() == eventingStateResumed {
+	if plan.State.ValueString() == eventingStateDeployed ||
+		plan.State.ValueString() == eventingStateResumed ||
+		!eventingFunctionHasChanged {
 		return
 	}
 
--- a/internal/resources/eventing_function.go
+++ b/internal/resources/eventing_function.go
@@ -172,9 +172,7 @@ func nullKeyspaceComputedAttributes(k *providerschema.EventingFunctionKeyspace) 
 	k.Collection = types.StringNull()
 }
 
-// eventingValueChanged reports whether a planned leaf represents a user-driven change: it holds a
-// concrete value (not null, not unknown) that differs from the prior state value. Null and unknown
-// planned values are ignored so computed attributes the user omitted never count as a change.
+// eventingValueChanged determines if scalar values are equal.
 func eventingValueChanged(plan, state attr.Value) bool {
 	if plan.IsNull() || plan.IsUnknown() {
 		return false
@@ -182,14 +180,10 @@ func eventingValueChanged(plan, state attr.Value) bool {
 	return !plan.Equal(state)
 }
 
-// eventingSettingsChanged reports whether the plan changes any settings value relative to the prior
-// state. A nil plan means the settings block was omitted and is not a user change.
+// eventingSettingsChanged determines if any of the eventing function settings have changed.
 func eventingSettingsChanged(plan, state *providerschema.EventingFunctionSettings) bool {
-	if plan == nil {
+	if plan == nil || state == nil {
 		return false
-	}
-	if state == nil {
-		state = &providerschema.EventingFunctionSettings{}
 	}
 
 	return eventingValueChanged(plan.WorkerCount, state.WorkerCount) ||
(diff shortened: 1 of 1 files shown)


internal/resources/eventing_function.go
```diff
@@ -346,14 +375,13 @@ func (e *EventingFunction) Update(ctx context.Context, req resource.UpdateReques
...
 	// change up front — before applying any activation state change. This prevents "inconsistent result after apply"
 	// errors.
-	if plan.State.ValueString() == eventingStateDeployed &&
-		(plan.Code != state.Code || eventingSettingsChanged(plannedSettings, stateSettings) || eventingBindingsChanged(plan.Bindings, state.Bindings)) {
+	if plan.State.ValueString() == eventingStateDeployed && eventingFunctionChanged(&plan, &state, plannedSettings, stateSettings) {
 		resp.Diagnostics.AddError(
-			"Cannot change eventing function settings or bindings while deployed",
-			"Eventing function "+name+" must be undeployed or paused before its settings or bindings can be changed.",
+			"Cannot change eventing function while deployed",
+			"Eventing function "+name+" must be undeployed or paused it can be changed.",
```
IsaacLambat:
```suggestion
			"Eventing function "+name+" must be in an undeployed or paused state in order to be changed.",
```

internal/resources/eventing_function.go
```diff
@@ -346,14 +375,13 @@ func (e *EventingFunction) Update(ctx context.Context, req resource.UpdateReques
 		name           = IDs[providerschema.FunctionName]
 	)
 
-	// Code, settings and bindings can only be changed while the function is undeployed or paused. Reject the
+	// eventing function can only be changed while the function is undeployed or paused. Reject the
 	// change up front — before applying any activation state change. This prevents "inconsistent result after apply"
 	// errors.
-	if plan.State.ValueString() == eventingStateDeployed &&
-		(plan.Code != state.Code || eventingSettingsChanged(plannedSettings, stateSettings) || eventingBindingsChanged(plan.Bindings, state.Bindings)) {
+	if plan.State.ValueString() == eventingStateDeployed && eventingFunctionChanged(&plan, &state, plannedSettings, stateSettings) {
```
IsaacLambat: Maybe should be checking if state is not `undeployed` or `paused`?


```suggestion
	if (plan.State.ValueString() != eventingStateUndeployed && plan.State.ValueString() != eventingStatePaused) && eventingFunctionChanged(&plan, &state, plannedSettings, stateSettings) {
```

IsaacLambat: Just checking, is it intended behaviour that we leave it to the API to check that the function is in the right state for the intended update operations: 
i.e. to modify eventing source or metadata, the function must be undeployed. Everything else can be done while paused

internal/resources/eventing_function.go
```diff
@@ -268,6 +268,35 @@ func eventingURLAuthChanged(plan, state *providerschema.EventingFunctionURLBindi
...
+		eventingValueChanged(plan.Code, state.Code) ||
+		eventingKeyspaceChanged(plan.EventSource, state.EventSource) ||
+		eventingKeyspaceChanged(plan.EventMetadataStorage, state.EventMetadataStorage) ||
+		eventingSettingsChanged(plannedSettings, stateSettings) ||
+		eventingBindingsChanged(plan.Bindings, state.Bindings)
+}
+
+// eventingKeyspaceChanged reports whether the plan changes the keyspace relative to the prior state. A
+// nil plan keyspace means it was omitted and is not a user change; a nil prior state with a non-nil
+// plan is a change.
```
IsaacLambat: We return false if `state == nil` which doesn't line up with this comment

internal/resources/eventing_function.go
```diff
@@ -268,6 +268,35 @@ func eventingURLAuthChanged(plan, state *providerschema.EventingFunctionURLBindi
...
 }
 
+// eventingFunctionChanged reports whether the plan changes any part of the function definition — its
+// name, description, code, event source/metadata keyspaces, settings or bindings — relative to the
+// prior state.
+func eventingFunctionChanged(
+	plan, state *providerschema.EventingFunctionResource,
+	plannedSettings, stateSettings *providerschema.EventingFunctionSettings,
+) bool {
+	return eventingValueChanged(plan.Name, state.Name) ||
```
IsaacLambat: nit: Name is require replace so this is not needed

internal/resources/eventing_function.go
```diff
@@ -381,12 +409,6 @@ func (e *EventingFunction) Update(ctx context.Context, req resource.UpdateReques
 		return
 	}
 
-	// if code, settings or bindings have not changed, skip update step.
-	if !eventingSettingsChanged(plannedSettings, stateSettings) && !eventingBindingsChanged(plan.Bindings, state.Bindings) &&
-		plan.Code == state.Code {
-		return
-	}
```
IsaacLambat, line 388: I think something like this is still needed, as if the user is making an update to move the function from `deployed` to `paused` for example, we will still do a PUT even though no config has changed other then the activation state

Can maybe at a variable near the top of the function that is used on L381 and L408, which simply stores the value of the output of `eventingFunctionChanged`

IsaacLambat (commented):
Potentially something that can be included in this PR: should the `eventingURLAuthChanged` function check if the password/token has changed in the plan compared to the state?

internal/resources/eventing_function.go
l0n3star: it's for terraform, not the API.

suppose state is changed to deployed and you change the event source keyspace.  the state will be changed to deployed.  when terraform tries to change the eventing function, the server eventing API will return an error.

this means terraform resource did not reach the desired state ie the keyspace did not change.  so we are now in an inconsistent state and provider will error out with `inconsistent result after apply`.

internal/resources/eventing_function.go
l0n3star: good catch

internal/resources/eventing_function.go
l0n3star: > we will still do a PUT even though no config has changed other then the activation state

sorry I'm confused with this comment. that is the reason I removed those lines.  

in other words, if I add those lines back, and only activation state has changed, then PUT won't be sent.

internal/resources/eventing_function.go
l0n3star: addressed in https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/664/commits/86c39c8dc592b640a71366c939e41cf9f0177b80

internal/resources/eventing_function.go
l0n3star: addressed in https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/664/commits/86c39c8dc592b640a71366c939e41cf9f0177b80

internal/resources/eventing_function.go
l0n3star: addressed in https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/664/commits/86c39c8dc592b640a71366c939e41cf9f0177b80

internal/resources/eventing_function.go
l0n3star: addressed in https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/664/commits/86c39c8dc592b640a71366c939e41cf9f0177b80

l0n3star:
> Potentially something that can be included in this PR: should the `eventingURLAuthChanged` function check if the password/token has changed in the plan compared to the state?

spot on

addressed in https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/664/commits/c5c4e7ab79068e419036ed3eaf55917887f88993

internal/resources/eventing_function.go
IsaacLambat: That sounds like behaviour we want.

If only the activation state has changed, then we should only set the activation state and not send a PUT.

If only other fields have changed (such as the code) then we can do the PUT.

If both the activation state and other fields have changed, then both requests can be made.

Why make the PUT call if only the state has changed?

l0n3star: apologies I misunderstood.  you're right.

addressed in https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/664/commits/9f81348d09c65f3578024e979abaac824571626b

IsaacLambat (approved):

PR #668, opened 2026-06-25
[AV-135261] Set url auth attribute to optional and computed

l0n3star:
## Jira

* AV-135261

## Description

if auth is omitted for url binding, it defaults to none.  so make this attribute both optional and computed.

will add acceptance tests in another PR

## Type of Change

- [X] Manually tested
- [ ] Unit tested
- [ ] Acceptance tested

### Testing

<details open>
  <summary>Testing</summary>

  see ticket for validation

</details>

## Further comments

### DIFF
--- a/internal/resources/eventing_function.go
+++ b/internal/resources/eventing_function.go
@@ -375,10 +375,11 @@ func (e *EventingFunction) Update(ctx context.Context, req resource.UpdateReques
 		name           = IDs[providerschema.FunctionName]
 	)
 
+	eventingFunctionHasChanged := eventingFunctionChanged(&plan, &state, plannedSettings, stateSettings)
+
 	// eventing function can only be changed while the function is undeployed or paused.
 	// in other states block any changes up front to prevent "inconsistent result after apply" errors.
-	if (plan.State.ValueString() != eventingStateUndeployed && plan.State.ValueString() != eventingStatePaused) &&
-		eventingFunctionChanged(&plan, &state, plannedSettings, stateSettings) {
+	if (plan.State.ValueString() != eventingStateUndeployed && plan.State.ValueString() != eventingStatePaused) && eventingFunctionHasChanged {
 		resp.Diagnostics.AddError(
 			"Cannot change eventing function while deployed",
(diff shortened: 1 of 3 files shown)


internal/resources/eventing_function.go
```diff
@@ -206,15 +234,15 @@ func eventingSettingsChanged(plan, state *providerschema.EventingFunctionSetting
 
 // eventingBindingsChanged determines if any of the bindings have changed,
 // except for secrets (password or bearer token).
-func eventingBindingsChanged(plan, state *providerschema.EventingFunctionBindingsResource) bool {
+func eventingBindingsChanged(ctx context.Context, plan, state *providerschema.EventingFunctionBindingsResource) (bool, error) {
 	if plan == nil || state == nil {
```
IsaacLambat: Out of interest, could this cause a problem if the user never had any bindings before, but now they are adding them? I assume in this case, this function would return false causing the users bindings noot to be updated.

IsaacLambat (commented):
Everything looks good, just 1 question

internal/resources/eventing_function.go
l0n3star: good one.  addressed in https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/668/commits/d648feb96463779a394117d15255d3535944611e

also tested that this works:
https://jira.issues.couchbase.com/browse/AV-135261?focusedCommentId=1430715

l0n3star: will have to check the rest of the resource but will address (if needed) in another PR

internal/resources/eventing_function.go
IsaacLambat, line 297: Same issue here as before?

IsaacLambat (approved):

internal/resources/eventing_function.go
l0n3star: will address in a follow up PR

l0n3star: addressed in https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/683

PR #686, opened 2026-07-06
[AV-134985] Add support for INCLUDE MISSING modifier in index key parsing

stanleefdz:
## Jira

* AV-134985

## Description

During terraform import, the provider now fetches the full CREATE INDEX DDL statement from the list definitions endpoint and parses the index key expressions from it, preserving modifiers like INCLUDE MISSING that are absent from the SecExprs field returned by the single-index API endpoint. If the DDL fetch or parse fails, it gracefully falls back to the original SecExprs behavior, ensuring backward compatibility.

## Type of Change

- [x] Manually tested
- [x] Unit tested
- [ ] Acceptance tested

### Testing

<details open>
  <summary>Testing</summary>
  
</details>

## Further comments

### DIFF
--- a/internal/resources/gsi.go
+++ b/internal/resources/gsi.go
@@ -406,24 +406,7 @@ func (g *GSI) Read(ctx context.Context, req resource.ReadRequest, resp *resource
 		state.IndexName = types.StringValue(indexName)
 		state.Status = types.StringValue(index.Status)
 
-		// Fetch the full DDL definition to get index keys with modifiers
-		// like INCLUDE MISSING, which are not present in SecExprs.
-		indexKeys := index.SecExprs
-		definition, err := g.getIndexDefinition(
-			ctx,
-			organizationID,
-			projectID,
-			clusterID,
-			bucketName,
-			scopeName,
-			collectionName,
-			indexName,
-		)
-		if err == nil && definition != "" {
-			if parsed, parseErr := parseIndexKeysFromDefinition(definition); parseErr == nil && len(parsed) == len(indexKeys) {
-				indexKeys = parsed
-			}
-		}
+		indexKeys := g.resolveIndexKeys(ctx, index.SecExprs, organizationID, projectID, clusterID, bucketName, scopeName, collectionName, indexName)
 
 		var keys []attr.Value
 		for _, key := range indexKeys {
@@ -816,6 +799,49 @@ func (g *GSI) getIndexDefinition(
 	return "", fmt.Errorf("index definition not found for %s", indexName)
 }
 
+// resolveIndexKeys attempts to enrich SecExprs with key modifiers (INCLUDE MISSING,
+// DESC, ASC) by fetching and parsing the full DDL definition. If any step fails it
+// logs a warning and returns the original SecExprs unchanged.
+func (g *GSI) resolveIndexKeys(
+	ctx context.Context, secExprs []string, organizationID, projectID, clusterID, bucketName, scopeName, collectionName, indexName string,
+) []string {
+	definition, err := g.getIndexDefinition(ctx, organizationID, projectID, clusterID, bucketName, scopeName, collectionName, indexName)
+	if err != nil {
+		tflog.Warn(ctx, "failed to fetch index DDL definition, falling back to SecExprs", map[string]any{
+			"index": indexName,
+			"error": err.Error(),
+		})
+		return secExprs
+	}
+
+	if definition == "" {
+		tflog.Warn(ctx, "index DDL definition was empty, falling back to SecExprs", map[string]any{
+			"index": indexName,
+		})
+		return secExprs
+	}
+
+	parsed, parseErr := parseIndexKeysFromDefinition(definition)
+	if parseErr != nil {
+		tflog.Warn(ctx, "failed to parse index keys from DDL definition, falling back to SecExprs", map[string]any{
+			"index": indexName,
+			"error": parseErr.Error(),
+		})
+		return secExprs
+	}
+
+	if len(parsed) != len(secExprs) {
+		tflog.Warn(ctx, "parsed key count does not match SecExprs count, falling back to SecExprs", map[string]any{
+			"index":          indexName,
+			"parsed_count":   len(parsed),
+			"secexprs_count": len(secExprs),
+		})
+		return secExprs
+	}
+
+	return mergeModifiers(secExprs, parsed)
+}
+
 // parseIndexKeysFromDefinition extracts index key expressions from a CREATE INDEX
 // DDL statement. Unlike SecExprs from the API, the DDL includes key modifiers
 // such as INCLUDE MISSING, DESC, and ASC.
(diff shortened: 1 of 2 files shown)


internal/resources/gsi.go
```diff
@@ -749,3 +768,125 @@ func (g *GSI) getQueryIndex(
...
+// DDL statement. Unlike SecExprs from the API, the DDL includes key modifiers
+// such as INCLUDE MISSING, DESC, and ASC.
+func parseIndexKeysFromDefinition(definition string) ([]string, error) {
+	// Find the key list which starts after ON <keyspace>( in the DDL.
+	// The keyspace reference (e.g., `bucket`.`scope`.`collection`) does not
+	// contain parentheses, so the first '(' after ON marks the key list start.
+	onIdx := strings.Index(strings.ToUpper(definition), " ON ")
+	if onIdx == -1 {
+		return nil, fmt.Errorf("could not find ON clause in definition")
+	}
```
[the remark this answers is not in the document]
stanleefdz: fixed it in [fac5d7f](https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/686/commits/fac5d7fbb21669c4ecb1e9f9268ccc0642857dac)

internal/resources/gsi.go
```diff
@@ -749,3 +768,125 @@ func (g *GSI) getQueryIndex(
...
+	uri := fmt.Sprintf(
+		"%s/v4/organizations/%s/projects/%s/clusters/%s/queryService/indexes?bucket=%s&scope=%s&collection=%s",
+		g.HostURL,
+		organizationID,
+		projectID,
+		clusterID,
+		bucketName,
+		scopeName,
+		collectionName,
+	)
```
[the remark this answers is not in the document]
stanleefdz: fixed it in [fac5d7f](https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/686/commits/fac5d7fbb21669c4ecb1e9f9268ccc0642857dac)

internal/resources/gsi_test.go
```diff
@@ -0,0 +1,102 @@
...
+		{
+			name:       "missing ON clause",
+			definition: "DROP INDEX `idx` ON `bucket`",
+			wantErr:    true,
+		},
+		{
+			name:       "primary index no keys",
+			definition: "CREATE PRIMARY INDEX `primary_idx` ON `bucket`.`scope`.`collection`",
+			wantErr:    true,
+		},
```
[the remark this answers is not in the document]
stanleefdz: fixed it in [fac5d7f](https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/686/commits/fac5d7fbb21669c4ecb1e9f9268ccc0642857dac)

internal/resources/gsi.go
```diff
@@ -406,8 +406,27 @@ func (g *GSI) Read(ctx context.Context, req resource.ReadRequest, resp *resource
 		state.IndexName = types.StringValue(indexName)
 		state.Status = types.StringValue(index.Status)
 
+		// Fetch the full DDL definition to get index keys with modifiers
```
PaulomeeCb: Do SecExprs and the DDL format keys the same way (backticks, spacing)? If not, could replacing all keys with the parsed DDL keys cause a new diff on a normal import, also, would it be safer to just append the modifier (like INCLUDE MISSING) onto the existing SecExprs value?

internal/resources/gsi.go
```diff
@@ -406,8 +406,27 @@ func (g *GSI) Read(ctx context.Context, req resource.ReadRequest, resp *resource
...
+		indexKeys := index.SecExprs
+		definition, err := g.getIndexDefinition(
+			ctx,
+			organizationID,
+			projectID,
+			clusterID,
+			bucketName,
+			scopeName,
+			collectionName,
+			indexName,
```
PaulomeeCb: If the DDL fetch or parse fails, the code falls back silently with no log. Please add a tflog.Debug/Warn saying it fell back to SecExprs, so it's easy to debug if an unexpected replacement happens.

stanleefdz: fixed it in [4cef6e4](https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/686/commits/4cef6e4d466a4befce9dc5bd0fc0c9f8d25ad77a)

internal/resources/gsi.go
stanleefdz: fixed it in [4cef6e4](https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/686/commits/4cef6e4d466a4befce9dc5bd0fc0c9f8d25ad77a)

l0n3star (changes requested):
good effort on this.  for some reason the parser on indexer is not extracting attributes for index keys.

however we should try to avoid writing our own parser.  this can be brittle.  for example there is vector attribute which is not handled in the PR.  in future there may be others.

instead we should just get this from query service, which correctly parses the index DDL.  here is an example:

```
SELECT i.*
FROM system:indexes i
WHERE bucket_id="source"
    AND scope_id="s1"
    AND keyspace_id="c1"
    AND name="user_notification_oktaId"
```

```
[
  {
    "bucket_id": "source",
    "datastore_id": "http://127.0.0.1:8091",
    "id": "3262e7fc30a43366",
    "index_key": [
      "`oktaId` INCLUDE MISSING"
    ],
    "keyspace_id": "c1",
    "metadata": {
      "definition": "CREATE INDEX `user_notification_oktaId` ON `source`.`s1`.`c1`(`oktaId` INCLUDE MISSING) WITH {  \"num_replica\":1 }",
      "last_scan_time": null,
      "num_replica": 1,
      "stats": {
        "last_known_scan_time": 0
      }
    },
    "name": "user_notification_oktaId",
    "namespace": "default",
    "namespace_id": "default",
    "scope_id": "s1",
    "state": "online",
    "using": "gsi",
    "with": {
      "num_partition": 1,
      "num_replica": 1,
      "retain_deleted_xattr": false
    }
  }
]
```

notice `index_key` has `INCLUDE MISSING`

stanleefdz:
@l0n3star your suggested approach is the best way to go about this, if it was actually possible to do so. The system:indexes approach is confirmed dead-on-arrival: isIndexStatement in couchbase-cloud/internal/cb/indexes.go rejects anything that isn't CREATE/DROP/ALTER/BUILD ...

I have fixed it with a different approach. During import, we fetch the index's CREATE INDEX DDL (which the server generates from its parsed AST and always includes key modifiers) and extract the key expressions verbatim by locating the parenthesized key list and splitting on top-level commas — no modifier interpretation at all, so INCLUDE MISSING, VECTOR, and any future modifiers pass through unchanged, with fallback to SecExprs if anything fails.

l0n3star:
I'm sorry this approach doesn't work either.  for example this is not parsed correctly:

```
CREATE INDEX `idx_composite` ON `default`([`type`, `name`],`country`) WITH {  \"num_replica\":1 }
```

please do not try to fix this.  there may be more bugs that I didn't find and we will waste too much time trying to resolve them.

it is a lot simpler just to have the server parse the DDL.  

> isIndexStatement in couchbase-cloud/internal/cb/indexes.go rejects anything that isn't CREATE/DROP/ALTER/BUILD

yes good catch.  what you can do is add another regex to allow querying system index table.  yeah I realize this means creating fixes on control plane and terraform, but it is a more stable solution.

PR #718, opened 2026-07-24
[AV-138196] Fix private_endpoints data source crash on non-empty list

nimiyajoseph:
## Jira

* AV-138196

## Description

Fixes a crash in the couchbase-capella_private_endpoints data source: reading it against a cluster that has one or more private endpoints failed with a Value Conversion Error. Adds a unit regression test and a real end-to-end acceptance test that exercises the previously-uncovered non-empty path, plus an opt-in flag to skip the shared app-service setup for focused runs.

The data source's nested data object declared 7 attributes (id, organization_id, project_id, cluster_id, cloud_provider, status, service_name), but the backing PrivateEndpointData struct defines only 3 (id, status, service_name). terraform-plugin-framework requires exact struct↔schema correspondence, so resp.State.Set failed whenever the endpoint list was non-empty:

```
Error: Value Conversion Error
Mismatch between struct and object type: Object defines fields not found in
struct: cloud_provider, organization_id, project_id, and cluster_id.
```

An empty list never triggers the element conversion, and the only existing test asserts data.# = 0, so this shipped undetected. It's directly on the critical path for automated PrivateLink: listing endpoints is how you discover the pending endpoint_id to accept.

The list-endpoints API returns only id/status/serviceName per endpoint (api.GetPrivateEndpointResponse and the generated PrivateEndpoint model), which matches the struct and the documented example output — only the schema was wrong.

**Changes**

- Fix: trim the nested data schema to id, status, service_name so it matches the struct and the API contract (internal/datasources/private_endpoints_schema.go).
- Docs: regenerate the nested schema in docs/data-sources/private_endpoints.md.
- Unit regression test: TestPrivateEndpointsSchemaMatchesStructNonEmpty sets a non-empty list and fails on the old schema, passes on the new one (internal/datasources/private_endpoints_test.go).
- End-to-end acceptance test: TestAccPrivateEndpointsDataSourceWithEndpoint — enables the service, creates a real AWS interface VPC endpoint against the exposed service_name, waits for the connection to register, reads the data source (non-empty list), and accepts the endpoint. Region-only aws provider (credentials come from the environment). Gated on ACC_AWS_REGION/ACC_AWS_VPC_ID/ACC_AWS_SUBNET_IDS/ACC_AWS_VPC_CIDR; skips cleanly when unset.
- Harness: ACC_SKIP_APP_SERVICE lets TestMain skip the shared app-service/app-endpoint provisioning for runs that don't need it (project/cluster/buckets still set up).

## Type of Change

- [x] Manually tested
- [ ] Unit tested
- [ ] Acceptance tested

### Testing

 JENKINS RUN: http://qe-jenkins1.sc.couchbase.com/job/terraform-test/1553/console - PASSED

## Further comments


internal/datasources/private_endpoints_schema.go
```diff
@@ -17,12 +17,13 @@ func PrivateEndpointsSchema() schema.Schema {
 	capellaschema.AddAttr(attrs, "cluster_id", privateEndpointsBuilder, requiredString())
 	capellaschema.AddAttr(attrs, "private_endpoint_dns", privateEndpointsBuilder, computedString(), "GetPrivateEndpointsResponse")
 
+	// The list-endpoints API returns only id, status and serviceName per
+	// endpoint (see api.GetPrivateEndpointResponse). The nested attributes must
+	// match the PrivateEndpointData struct exactly or terraform-plugin-framework
+	// fails state conversion whenever the list is non-empty; org/project/cluster
+	// already exist at the top level and cloud_provider is not returned here.
```
IsaacLambat: I don't think we really need a comment regarding the context of this bug to live in the code, just seems unnecessary as it wouldn't help anyone in the future

internal/datasources/private_endpoints_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+	providerschema "github.com/couchbasecloud/terraform-provider-couchbase-capella/internal/schema"
+)
+
+// TestPrivateEndpointsSchemaMatchesStructNonEmpty guards against a struct/schema
+// mismatch on the couchbase-capella_private_endpoints data source. The nested
+// "data" attributes must line up exactly with the PrivateEndpointData struct, or
+// terraform-plugin-framework fails state conversion the moment the endpoint list
+// is non-empty (the empty-list path never exercises the mismatch, which is why
+// the acceptance test that only checks data.# = 0 did not catch it).
```
IsaacLambat: unnecessary context of this bug that will be irrelevant when this pr is merged

acceptance_tests/private_endpoint_acceptance_test.go
```diff
@@ -204,3 +208,207 @@ resource "couchbase-capella_private_endpoints" "%[2]s" {
...
+					globalOrgId, globalProjectId, globalClusterId,
+				),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr(serviceRef, "enabled", "true"),
+					resource.TestCheckResourceAttrSet(serviceRef, "service_name"),
+					resource.TestCheckResourceAttrSet(vpceRef, "id"),
+					resource.TestCheckResourceAttrSet(dsRef, "private_endpoint_dns"),
+					// Regression guard: converting a NON-EMPTY endpoint list to
+					// state must succeed. Pre-fix this step errored with a "Value
+					// Conversion Error" before any check below could run.
```
IsaacLambat: I don't think we need the pre-fix part, given its implied that it used to just error. Plus we have the regression bug details on L218
```suggestion
					// Regression guard: converting a NON-EMPTY endpoint list to state must succeed. 
```

acceptance_tests/private_endpoint_acceptance_test.go
```diff
@@ -204,3 +208,207 @@ resource "couchbase-capella_private_endpoints" "%[2]s" {
...
+//
+// The AWS credentials come from the ambient environment (the aws provider is
+// region-only). In cross-account CI, assume the target-account role in the
+// pipeline and export AWS_ACCESS_KEY_ID/SECRET/SESSION_TOKEN before running.
+func TestAccPrivateEndpointsDataSourceWithEndpoint(t *testing.T) {
+	region := os.Getenv("ACC_AWS_REGION")
+	vpcID := os.Getenv("ACC_AWS_VPC_ID")
+	// Parse up front so a value that is only commas/whitespace (which would
+	// render to subnet_ids = [] and fail later with a confusing AWS error) is
+	// caught here as "no subnets" rather than passing a bare non-empty check.
```
IsaacLambat: It makes sense to parse subnet IDs, so this comment seems unnecessary 

Plus theres a godoc on the `parseSubnetIDs` function anyway

acceptance_tests/private_endpoint_acceptance_test.go
```diff
@@ -204,3 +208,207 @@ resource "couchbase-capella_private_endpoints" "%[2]s" {
...
+// the Capella cluster:
+//
+//	ACC_AWS_REGION      e.g. us-east-1 (must match the cluster region)
+//	ACC_AWS_VPC_ID      e.g. vpc-0123456789abcdef0
+//	ACC_AWS_SUBNET_IDS  comma-separated, e.g. subnet-aaa,subnet-bbb
+//	ACC_AWS_VPC_CIDR    e.g. 10.0.0.0/16
+//
+// The AWS credentials come from the ambient environment (the aws provider is
+// region-only). In cross-account CI, assume the target-account role in the
+// pipeline and export AWS_ACCESS_KEY_ID/SECRET/SESSION_TOKEN before running.
```
IsaacLambat: How does this test work on CI? No CI changes have been made to support it so a bit confused as to the purpose of this comment?

IsaacLambat (commented):
Fix looks good. Just a lot of unnecessary comments

acceptance_tests/private_endpoint_acceptance_test.go
nimiyajoseph: We plan to skip this for GitHub Workflow; as of now only running this test as part of the QE pipeline http://qe-jenkins1.sc.couchbase.com/job/terraform-test/1553/console so that we can use QE AWS creds in the pipeline
To run as part of GitHub Workflow, we need to create a new AWS account specific to TF, so skipping for now

acceptance_tests/private_endpoint_acceptance_test.go
nimiyajoseph: makes sense , i will remove it

acceptance_tests/private_endpoint_acceptance_test.go
nimiyajoseph: Will remove the error part

internal/datasources/private_endpoints_schema.go
nimiyajoseph: i will remove this too

internal/datasources/private_endpoints_test.go
nimiyajoseph: will remove all unnecessary comments

IsaacLambat (approved):

PR #720, opened 2026-07-27
[AV-138151] Bound API client retries with exponential backoff and jitter

PaulomeeCb:
## Jira

* [AV-138151](https://couchbasecloud.atlassian.net/browse/AV-138151)

## Description

The shared API client's `exec` retry loop (`internal/api/client.go`) had two problems for 503 and 504 responses:

1. **No attempt cap** — retries were only bounded by the 10-minute context timeout, so a persistently-down endpoint was retried ~300 times over the full 10 minutes before returning a generic deadline error.
2. **Fixed 2-second wait with no backoff or jitter** — every retry waited exactly 2 seconds, hammering a struggling backend at a flat cadence. Only 429 (rate-limit) had proper backoff via its mandatory `Retry-After` header.

**Fix:**

- Added `maxRetryAttempts = 5` to cap retries for 503, 504, and 429.
- Introduced exponential backoff with equal jitter (`2s → 4s → 8s → 16s → 30s`), so a call fails in ~1 minute instead of ~10 minutes.
- `Retry-After` headers (from 429 or 503/504) are still honored and take precedence over computed backoff.
- On exhaustion, returns a typed `*RetriesExhaustedError` wrapping the original sentinel, making it inspectable via `errors.Is` and `errors.As`.

> **Fix confidence:** 95%
> **Reviewer action required:** The ticket flags whether non-idempotent POST should be retried on 5xx at all. The current logic retries 503/504 for all HTTP methods (including POST). This PR does not change that; please review and decide whether a separate follow-up should make retries conditional on method/idempotency.

## Type of Change

### How was this change tested and do you have evidence?

- [x] Unit tested

### Testing

<details open>
  <summary>Testing</summary>

Added 11 table-driven unit tests in `internal/api/client_test.go` asserting:

- (a) Attempt cap is enforced for 503, 504, and 429.
- (b) Exponential backoff grows monotonically between attempts; is capped at 30s.
- (c) Jitter produces varying values across runs within the expected bounds.
- (d) `Retry-After` is honored when the function returns a non-zero duration.
- (e) Exhaustion returns a typed `*RetriesExhaustedError` wrapping the original sentinel.
- (f) Non-retryable errors stop the loop immediately.
- (g) Context cancellation works correctly.
- (h) `isRetryable` correctly classifies sentinel errors.

All tests pass; `go vet`, `goimports`, `make fmt`, and `go build` all clean.
Acceptance tests compile-checked with `go test -c -o /dev/null ./acceptance_tests/`.

</details>

## Further comments

Generated by the tf-ticket-autofix droid. Draft — a human reviewer makes the merge decision.

[AV-138151]: https://couchbasecloud.atlassian.net/browse/AV-138151?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/api/client.go
+++ b/internal/api/client.go
@@ -7,6 +7,7 @@ import (
 	goer "errors"
 	"fmt"
 	"io"
+	"math/rand"
 	"net/http"
 	"strconv"
 	"time"
@@ -56,6 +57,15 @@ type EndpointCfg struct {
 
 // defaultWaitAttempt re-attempt http request after 2 seconds.
 const defaultWaitAttempt = time.Second * 2
+
+// maxRetryAttempts caps the number of retries for retryable errors
+// (rate limit, service unavailable, gateway timeout) so that a
+// persistently unavailable endpoint fails fast instead of hanging
+// for the full context timeout (10 minutes).
+const maxRetryAttempts = 5
+
+// maxBackoff caps exponential backoff to prevent unbounded growth.
+const maxBackoff = 30 * time.Second
 
 // ExecuteWithRetry is used to construct and execute a HTTP request with retry.
 // It then returns the response.
@@ -182,6 +192,7 @@ func exec(
 		err      error
 		backOff  time.Duration
 		response *Response
+		attempts int
 	)
 
 	const timeout = time.Minute * 10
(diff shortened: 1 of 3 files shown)


PaulomeeCb:
🔄 **Duplicate droid invocation detected (2026-07-28).**

The tf-ticket-autofix droid was invoked again for AV-138151. This PR (#720) already addresses the ticket, so the droid short-circuited at Phase 1b (deduplication).

No action needed — this comment is informational only.

PR #730, opened 2026-07-29
[AV-137441] Add acceptance test for audit log settings disable on removal

panigrahisubhrajit:
## Jira

* AV-137441

## Description

Adds `TestAccAuditLogSettingsResourceDisableOnRemoval` acceptance test. When a user removes the `audit_log_settings` resource from their Terraform config and applies, the provider should revert audit logging to disabled (the default). Previously, removing the resource left audit logging enabled on the server, blocking cluster downgrade from Enterprise to Dev Pro. The API fix for the audit log disable endpoint has now landed, so this test runs and validates the end-to-end disable-on-removal behavior.

## Type of Change

- [x] Manually tested
- [x] Acceptance tested
- [ ] Unit tested

### Testing

<details open>
  <summary>Testing</summary>

1. Creates a cluster with `audit_log_settings` (audit enabled, 3 event IDs)
2. Removes the `audit_log_settings` resource (cluster-only config)
3. Hits the API directly to verify `auditEnabled=false` and no event IDs remain

Step 1 assertions include org/project IDs, cluster_id, specific enabled_event_ids (20488, 20490, 20491), and disabled_users count for thorough validation.

</details>

## Further comments

The v4 Audit Log PUT endpoint previously returned `{"code":10000,"httpStatusCode":500,"message":"An internal server error occurred."}` when the provider sent a request to disable audit logging. This server-side issue has been fixed.


acceptance_tests/audit_log_settings_acceptance_test.go
```diff
@@ -551,3 +551,105 @@ func testAccExistsAuditLogSettingsResource(t *testing.T, resourceReference strin
 		return err
 	}
 }
+
+func TestAccAuditLogSettingsResourceDisableOnRemoval(t *testing.T) {
+	t.Skip("CBSE-23063/AV-137441: API returns 500 when disabling audit log settings on removal; un-skip once the API fix lands")
```
nimiyajoseph: you can remove the skip as the issue is now fixed

acceptance_tests/audit_log_settings_acceptance_test.go
```diff
@@ -551,3 +551,105 @@ func testAccExistsAuditLogSettingsResource(t *testing.T, resourceReference strin
...
+
+	clusterResourceName := randomStringWithPrefix("tf_acc_audit_rm_cluster_")
```
nimiyajoseph, line 556: can't we use global cluster instead? Is it really required to have a separate cluster for this test alone?

panigrahisubhrajit: The existing audit tests also create their own cluster for the same isolation reason. A dedicated cluster is necessary here. if it runs against the global cluster, it would permanently break audit state for every other test that relies on the global cluster (e.g  TestAccAuditLogSettingsResource)

IsaacLambat: Can we please have a ticket to convert all the tests to use the global cluster (or have them all share their own cluster if needed)? We shouldn't be deploying a cluster per audit test.

Please look at how the App Service log streaming acceptance tests are done to allow sharing the same cluster, without tests conflicting with each other.

acceptance_tests/utils.go
```diff
@@ -33,25 +33,36 @@ func randomStringWithPrefix(prefix string) string {
...
+// private ranges to avoid conflicts with existing clusters in the shared test
+// organization. It randomly picks one of:
+//
+//	10.A.B.0/23    (A: 0-255, B: even 0-254)  → 32,768 blocks
+//	172.C.D.0/21   (C: 16-31, D: multiple of 8) → 512 blocks
+//	192.168.E.0/24 (E: 0-255)                  → 256 blocks
 func generateRandomCIDR() string {
 	// Use crypto/rand for cryptographically secure randomness
-	buf := make([]byte, 2)
+	buf := make([]byte, 3)
```
IsaacLambat, line 45: can this change be done as a separate ticket?

panigrahisubhrajit: this is a small change and as cluster deployment was failing i had to add it here.

IsaacLambat (approved):
Approved on the premise that these audit logging tests will all be refactored in the near future to use the same cluster (ideally the global one)

PR #731, opened 2026-07-29
[AV-137404] Add direct API test for api key invalid expiry validation

panigrahisubhrajit:
## Jira

* AV-137404

## Description

Adds `TestAccApiKeyResourceInvalidExpiryDirectAPI` acceptance test. When a customer creates an API key with `expiry=0`, the API should return a proper validation error (HTTP 422), not a generic 500 (code 10000). This test hits the v4 API directly and verifies the server behavior.

## Type of Change

- [ ] Manually tested
- [x] Acceptance tested
- [ ] Unit tested

### Testing

<details open>
  <summary>Testing</summary>

1. Sends `POST /v4/organizations/{orgId}/apikeys` with `expiry=0`
2. Verifies the API rejects the request (non-nil error)
3. Fails with a clear error if the API still returns HTTP 500 (regression)
4. Fails on unexpected success (API accepted invalid expiry)
5. Fails if the error is not an `api.Error` type (e.g., auth/network errors)

The API key name uses `randomStringWithPrefix` to avoid collisions. Error paths use `t.Fatal` for clean failure.

</details>

## Further comments

Test passes against the current tenant (v2.2.283+). If the tenant predates the AV-137404 fix, the test will fail with the CBSE-23222 regression message.


acceptance_tests/apikey_acceptance_test.go
```diff
@@ -495,3 +500,41 @@ func generateApiKeyImportIdForResource(resourceReference string) resource.Import
...
+// TestAccApiKeyResourceInvalidExpiryDirectAPI hits the API directly with an
+// invalid expiry (0) to verify CBSE-23222 / AV-137404: the API must return a
+// proper validation error (HTTP 422), not a generic 500/code 10000. Fails
+// against a tenant that predates the AV-137404 fix (v2.2.283); passes once the
+// tenant has it.
+func TestAccApiKeyResourceInvalidExpiryDirectAPI(t *testing.T) {
+	data := newTestClient(t)
+
+	name := randomStringWithPrefix("tf_acc_apikey_expiry_direct_")
+	expiry := float32(0)
```
nimiyajoseph, line 513: also add exact validation of 

```
{
    "code": 422,
    "hint": "Please review your request and ensure that all required parameters are correctly provided.",
    "httpStatusCode": 422,
    "message": "APIKey token expiry must be greater or equal to .01."
}
```

panigrahisubhrajit: Done

nimiyajoseph (dismissed):

acceptance_tests/apikey_acceptance_test.go
```diff
@@ -495,3 +500,68 @@ func generateApiKeyImportIdForResource(resourceReference string) resource.Import
 		return fmt.Sprintf("id=%s,organization_id=%s", rawState["id"], globalOrgId), nil
 	}
 }
+
+// TestAccApiKeyResourceInvalidExpiryDirectAPI hits the API directly with an
+// invalid expiry (0) to verify CBSE-23222 / AV-137404: the API must return a
+// proper validation error (HTTP 422) with a specific expiry message, not a
+// generic 500/code 10000. Fails against a tenant that predates the AV-137404
+// fix (v2.2.283); passes once the tenant has it.
```
IsaacLambat: When you say tenant here, do you mean the CP? If so, I don't think this is necessary to know at all

acceptance_tests/apikey_acceptance_test.go
```diff
@@ -495,3 +500,68 @@ func generateApiKeyImportIdForResource(resourceReference string) resource.Import
...
+//
+// Expected error response when the fix is present:
+//
+//	{
+//	  "code": 422,
+//	  "hint": "Please review your request and ensure that all required parameters are correctly provided.",
+//	  "httpStatusCode": 422,
+//	  "message": "APIKey token expiry must be greater or equal to .01."
+//	}
+func TestAccApiKeyResourceInvalidExpiryDirectAPI(t *testing.T) {
```
IsaacLambat, line 509: I am a bit confused as to why this test is going in the TF repo, given that it's not testing any TF? It looks like a test that belongs in the cloud repo

acceptance_tests/apikey_acceptance_test.go
```diff
@@ -495,3 +500,68 @@ func generateApiKeyImportIdForResource(resourceReference string) resource.Import
...
+	}
+
+	var apiErr *api.Error
+	if !errors.As(err, &apiErr) {
+		t.Fatalf("expected api.Error, got %T: %v", err, err)
+	}
+
+	// Exact validation of the expected error response:
+	//   {"code":422, "hint":"Please review your request ...", "httpStatusCode":422,
+	//    "message":"APIKey token expiry must be greater or equal to .01."}
```
IsaacLambat: This is already in the GoDoc for the test so we don't need this duplicated.

acceptance_tests/apikey_acceptance_test.go
```diff
@@ -495,3 +500,68 @@ func generateApiKeyImportIdForResource(resourceReference string) resource.Import
...
+				apiErr.Error(),
+			)
+		}
+		t.Fatalf(
+			"unexpected HTTP status for invalid expiry: got %d, want %d. Body: %s",
+			apiErr.HttpStatusCode, http.StatusUnprocessableEntity, apiErr.Error(),
+		)
+	}
+	if apiErr.Code != 422 {
+		t.Fatalf("unexpected error code: got %d, want 422. Body: %s", apiErr.Code, apiErr.Error())
```
IsaacLambat, line 547: This is never going to be hit as the Fatalf on L553 would have always been hit before this. I think the logic between L545 and L566 needs cleaning up as its a bit confusing and all over the place.

The testify assert/require package would be useful to use here

acceptance_tests/utils.go
```diff
@@ -33,25 +33,36 @@ func randomStringWithPrefix(prefix string) string {
 	return prefix + randomString()
 }
 
-// generateRandomCIDR generates a random /23 CIDR block in the 10.0.0.0/8 private range
-// to avoid conflicts with existing clusters in the organization.
-// Format: 10.X.Y.0/23 where X is 0-255 and Y is an even number (0, 2, 4, ..., 254)
+// generateRandomCIDR generates a random CIDR block across all three RFC 1918
```
IsaacLambat: This change is irrelevant to this PR.

IsaacLambat:
This change is not manually tested as it is an acceptance test, so the PR body can be adjusted to reflect this

acceptance_tests/apikey_acceptance_test.go
panigrahisubhrajit: Removed — the tenant/version reference has been dropped from the GoDoc.

acceptance_tests/apikey_acceptance_test.go
panigrahisubhrajit: This test guards against a server-side regression (CBSE-23222) that directly impacts the Terraform provider. When the provider sends an expiry it can't validate client side, the API must return a proper 422 otherwise the provider's error handling breaks.

Happy to move it to the cloud repo if you'd prefer, but having it here alongside the other API key acceptance tests keeps the guard close to the code that depends on it.

acceptance_tests/apikey_acceptance_test.go
panigrahisubhrajit: Removed , the duplicate inline comment and the redundant JSON block in GoDoc have been dropped.

acceptance_tests/apikey_acceptance_test.go
panigrahisubhrajit: Fixed, the nested if/else has been flattened to sequential checks. 500 regression first (bail early), then 422 validation, then code/message/hint. No dead code paths.

acceptance_tests/utils.go
panigrahisubhrajit: Removed — the generateRandomCIDR change has been reverted. The diff now only contains the test file changes relevant to this PR.

acceptance_tests/apikey_acceptance_test.go
```diff
@@ -539,17 +530,14 @@ func TestAccApiKeyResourceInvalidExpiryDirectAPI(t *testing.T) {
 		t.Fatalf("expected api.Error, got %T: %v", err, err)
 	}
 
-	// Exact validation of the expected error response:
-	//   {"code":422, "hint":"Please review your request ...", "httpStatusCode":422,
-	//    "message":"APIKey token expiry must be greater or equal to .01."}
+	if apiErr.HttpStatusCode == http.StatusInternalServerError {
+		t.Fatalf(
```
IsaacLambat, line 534: In future, please use the testify package to do these requires/asserts

acceptance_tests/apikey_acceptance_test.go
IsaacLambat: Would be good if we can have this in the cloud repo as well, if you do want to keep it here

IsaacLambat (approved):

acceptance_tests/apikey_acceptance_test.go
panigrahisubhrajit: Ok will do it in a separate ticket.

PR #737, opened 2026-08-03
[AV-138933] Add bucket update acceptance test and un-skip app service load balancer CIDR tests

nimiyajoseph:
## Jira

* [AV-138933](https://jira.issues.couchbase.com/browse/AV-138933)

## Description

**1. App Service `load_balancer_cidr` tests now actually run.**

`app_service_load_balancer_cidr_acceptance_test.go` previously contained three tests, all of which were skipped or gated, so the feature had effectively no automated coverage. The file is now a single unskipped test that provisions its own Azure cluster (the attribute is Azure-only, so the shared AWS cluster cannot be used) and reuses that one cluster across all steps:

| Step | Assertion |
|---|---|
| 1–5 | Malformed CIDRs — create must fail |
| 6 | Create — attributes, plus `app_services` data source round-trip |
| 7 | Import — `ImportStateVerify` |
| 8 | Scale nodes 2→3 — `ExpectResourceAction(Update)` |
| 9 | Change CIDR — `ExpectResourceAction(Replace)` |

Rejection cases run first, while no app service exists, so each is a clean create failure rather than a replace-then-fail. Step 6 revives `testAccCheckAppServicesDataSourceLoadBalancerCIDR`, which was previously dead code inside a skipped test — it is the only check that the org-wide list endpoint and the single-resource GET agree.

**2. New bucket update coverage**

`bucket_update_acceptance_test.go` creates a bucket with only `memory_allocation_in_mb` set, then resizes it. `Bucket.Update` builds the PUT body from the plan using `ValueString()`/`ValueInt64()`, so any Optional+Computed attribute the practitioner omitted arrives unknown and is transmitted as `""` or `0`. The API rejects `durabilityLevel: ""` with a 422; `replicas: 0` and `timeToLiveInSeconds: 0` would apply silently.

**3. Regression guards for [AV-138536.](https://jira.issues.couchbase.com/browse/AV-138536)**

`cluster_acceptance_test.go` adds `ExpectResourceAction(Update)` to the GCP horizontal-scaling step. That config omits `couchbase_server`, which is Optional+Computed with no `Default`, so it is marked unknown on every update plan and is safe only because it also carries `UseStateForUnknown`. Drop that modifier and the scale silently becomes a cluster destroy-and-recreate. The step already performs an in-place scale, so the assertion costs nothing.

`appservice_acceptance_test.go` makes the unsupported-version step `PlanOnly`. An apply-time assertion also passes when Terraform recreates the app service instead of updating it — in which case the API rejects the version and the provider's own guard is never exercised.

## Type of Change

- [ ] Manually tested
- [ ] Unit tested
- [x] Acceptance tested

### Testing

<details open>
  <summary>Testing</summary>
  
</details>

## Further comments

[AV-138933]: https://couchbasecloud.atlassian.net/browse/AV-138933?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


stanleefdz (approved):
LGTM

PR #742, opened 2026-08-04
[AV-139105] Add sanity test suite as the pull request gate

nimiyajoseph:
## Jira

* AV-139105

## Description

Makes the P0 sanity suite the acceptance-test gate for every pull request, and fills out that suite so it actually covers the provider.

Previously `acceptance-tests.yml` ran the **entire** acceptance suite on every PR (`make testacc`, `-timeout=180m`). That is far too slow and too resource-hungry to be a useful gate, and it provisions dozens of real clusters and app services per PR. Meanwhile `acceptance_tests/sanity.list` existed but was not wired into CI at all, and covered only 18 tests.

Three changes:

**1. Expanded the sanity list from 18 to 77 tests.** P0 is defined here as the primary happy-path (create / read / update / import / delete) test for every resource and data source the provider exposes. Negative and validation-only tests (invalid UUIDs, missing required fields, bad enums) are P1+ and stay out, as do tests that unconditionally `t.Skip`. Newly covered areas that previously had *zero* sanity coverage: buckets / scopes / collections (only `TestAccSampleBucket` was there, not the actual bucket resource), free tier clusters, cluster on/off and deletion protection, eventing, audit logging, Data API, events, cluster stats, certificates, on-demand backup and backup schedules, snapshot restore data sources, and the read/list data sources for every resource already in the list.

lso fixes a silently-dead entry: the list contained `TestAccPrivateEndpointServiceEnable`, which does not exist — the real test is `TestAccPrivateEndpointServiceEnableDisable`. Because `run-test-list.sh` anchors the pattern with `^(...)$`, that entry was matching nothing rather than erroring.

**2. Split the CI triggers.** New `sanity-tests.yml` runs the sanity suite on every `pull_request`, with no paths filter. `acceptance-tests.yml` drops its `pull_request` trigger and now runs the full suite on push-to-main plus `workflow_dispatch`. Running both on PRs would have duplicated cost and let the two runs collide on shared test fixtures (`globalClusterId` and friends) in the same Capella organization.

**3. Made the runner usable.** Added `make testacc-sanity` and a generic `make testacc-list` (honors `TEST_LIST`) — note that `sanity.list` already documented `make testacc-list` in its header, but that target never existed. `run-test-list.sh` had a hardcoded `-timeout=30m`, which at 77 cluster-provisioning tests would have panicked on timeout instead of reporting test failures; it is now `${TEST_TIMEOUT:-180m}`, matching `make testacc`. The script also no longer clobbers a caller-supplied `CAPELLA_OPENAPI_SPEC_PATH`, so the workflow's `github.workspace` path survives.

## Type of Change

- [x] This change updates the ci/cd workflow.

- [ ] Manually tested
- [ ] Unit tested
- [x] Acceptance tested

### Testing

<details open>
  <summary>Testing</summary>
  
</details>

## Further comments

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -1,9 +1,5 @@
 name: Acceptance Tests
 
-# The full acceptance suite. Pull requests are gated by the P0 subset in
-# sanity-tests.yml instead: running both against the same Capella organization
-# duplicates cost and lets the two runs collide on shared test fixtures.
-# Use workflow_dispatch to run the full suite against a branch on demand.
 on:
   workflow_dispatch:
   push:
--- a/.github/workflows/sanity-tests.yml
+++ b/.github/workflows/sanity-tests.yml
@@ -1,22 +1,27 @@
-name: Acceptance Tests
+name: Sanity Tests
 
-# The full acceptance suite. Pull requests are gated by the P0 subset in
-# sanity-tests.yml instead: running both against the same Capella organization
-# duplicates cost and lets the two runs collide on shared test fixtures.
-# Use workflow_dispatch to run the full suite against a branch on demand.
+# Runs the P0 acceptance test suite (acceptance_tests/sanity.list) as the
+# pull request gate. The full acceptance suite lives in acceptance-tests.yml.
+# No paths filter: the sanity suite gates every pull request.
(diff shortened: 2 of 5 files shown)


stanleefdz (approved):
LGTM

PR #751, opened 2026-08-14
[AV-135211] Recognize "Created" as a deferred index build status

panigrahisubhrajit:
## Jira

* [AV-135211](https://couchbasecloud.atlassian.net/browse/AV-135211)

## Description

`couchbase-capella_query_index_monitor` polled to its own timeout (~60 min) on a `defer_build = true` index instead of returning immediately. The root cause: the V4 `queryService/indexBuildStatus` endpoint returns `"Created"`, not `"Deferred"`, for a deferred index, so `WatchIndexes` never matched its early-exit check (`internal/api/gsi.go`). `WatchIndexes` now treats `"Created"` the same as `"Deferred"` **only when the caller is waiting for `"Ready"`** — `WatchIndexes` is also called by the BUILD INDEX flow with `desiredState="Created"`, where reaching `"Created"` is success, not a stuck index, so the check has to be conditional on that. `TestAccDatasourceQueryIndexMonitorDeferred` is un-skipped now that the actual status value is confirmed.

Un-skipping that test added real concurrent `CREATE`/`DROP INDEX` load to the shared sanity-suite cluster for the first time, and CI runs on this PR surfaced three more pre-existing eventual-consistency gaps in the GSI resource's handling of the query service's distributed catalog — not new bugs introduced here, just latent ones this extra load finally triggered:

- `internal/resources/gsi.go` (`Read`): a fresh index's single-item lookup could transiently 404 right after `CREATE INDEX` under load, wiping it from state and causing a spurious "will be created" on the next `terraform plan`. Now retries a bounded few times before concluding the index is genuinely gone.
- `internal/resources/gsi.go` (`Delete`) / `internal/errors/errors.go`: `DROP INDEX` on an index still finishing its background build failed hard with a 500 ("scheduled for background creation"), and a 404 on an already-gone index also failed hard instead of being treated as a no-op. Added `ErrIndexDropPending`, retried with the same bounded backoff, and made not-found idempotent.
- `acceptance_tests/query_index_acceptance_test.go`: the `query_indexes` datasource's list endpoint lagged behind a freshly created index under load (a separate, less consistent endpoint than the single-item lookup above); split the affected tests into a create step + a datasource-read step with a short delay. Also split `TestAccDatasourceQueryIndexMonitorDeferred` so the deferred index is created in its own (non-erroring) step for reliable cleanup, and tightened its `ExpectError` regex to match terraform's word-wrapped diagnostic text and the specific `ErrIndexDeferred` message rather than a generic substring.

## Type of Change

- [x] Acceptance tested

### Testing

<details open>
  <summary>Testing</summary>

Confirmed against a live tenant locally (`TestAccGSI`, `TestAccDatasourceQueryIndexes`, `TestAccDatasourceQueryIndexMonitorDeferred`, `TestAccDatasourceQueryIndexMonitorReady` all PASS), and via the CI Sanity suite on this PR: `TestAccGSI` PASS, `TestAccDatasourceQueryIndexes` PASS, `TestAccDatasourceQueryIndexMonitorReady` PASS, `TestAccDatasourceQueryIndexMonitorDeferred` PASS. Remaining CI failures on this PR (`TestAccProjectResource`/`TestAccProjectCreateWithReqFields` on a tenant project-quota limit, `TestAccFlushBucketResource` on a raw API 500, `TestAccAppServiceLogStreaming/App_Endpoint_Logging_Config` on a timeout, `TestAccClusterResourceAzureDiskAutoExpansion` on a `deploymentFailed` cluster) are pre-existing shared-tenant flakiness unrelated to any file this PR touches.

</details>

## Further comments

[AV-135211]: https://couchbasecloud.atlassian.net/browse/AV-135211?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/acceptance_tests/query_index_acceptance_test.go
+++ b/acceptance_tests/query_index_acceptance_test.go
@@ -264,8 +264,10 @@ func TestAccDatasourceQueryIndexMonitorDeferred(t *testing.T) {
 				Config: testAccQueryIndexOnlyConfigWithDeferBuild(idxName, "deferred_field", true),
 			},
 			{
-				Config:      testAccQueryIndexMonitorDeferredConfig(idxName, monitorName),
-				ExpectError: regexp.MustCompile(`(?is)Error monitoring query indexes.*is in Deferred state and will not become Ready without a BUILD INDEX statement`),
+				Config: testAccQueryIndexMonitorDeferredConfig(idxName, monitorName),
+				// terraform's diagnostic renderer word-wraps the message at the terminal width, so
+				// whitespace between words (e.g. "BUILD INDEX\nstatement") can become a newline.
+				ExpectError: regexp.MustCompile(`(?is)Error\s+monitoring\s+query\s+indexes.*is\s+in\s+Deferred\s+state\s+and\s+will\s+not\s+become\s+Ready\s+without\s+a\s+BUILD\s+INDEX\s+statement`),
 			},
 		},
 	})
--- a/acceptance_tests/query_index_acceptance_test.go
+++ b/acceptance_tests/query_index_acceptance_test.go
@@ -228,10 +228,7 @@ func TestAccDatasourceQueryIndexMonitorMultipleIndexes(t *testing.T) {
 	})
 }
 
-// AV-132671: API returns 200 OK with non-"Deferred" status for a deferred index; WatchIndexes polls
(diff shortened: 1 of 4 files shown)


stanleefdz (approved):
LGTM

PR #752, opened 2026-08-18
[AV-139978] Use List types for access across credential, role and datasources

stanleefdz:
## Jira

* AV-139978

## Description

Fix Value Conversion Error on every database credential plan

Merging main into the fine-grained RBAC P3 branch resolved the access attribute toward List but left its validator slice as []validator.Set, which does not compile, and left ValidateConfig reading access into a types.Set. GetAttribute matches the target against the schema type at runtime, so once the validator slice is corrected the provider builds and vets cleanly while failing every database credential plan -- including basic credentials that never touch fine-grained RBAC.

ValidateConfig now reads access and user_roles into attr.Value. That target short-circuits the framework's type reflection, so the schema's declared type can no longer drift away from what the resource reads back. Only IsNull/IsUnknown are needed here and both are on the interface.

The same merge silently reverted AV-139841 in two places it did not name: database_role declared access, buckets, scopes and collections as Sets, and the new shared datasource helpers built Sets in place of the Lists main's credentials datasource published. Both are converted to Lists, with a computedStringList helper added alongside computedStringSet. privileges and user_roles stay Sets -- they are unordered name sets and benefit from order-insensitive diffs.

Also rejects `access = []`, which satisfied ExactlyOneOf but marshalled away under the omitempty tag, producing a request body carrying neither access nor userRoles.

Adds a plan-only acceptance test covering basic, advanced and empty access configurations, and regenerates docs -- including the previously missing database_role, database_roles and database_privileges pages.

## Type of Change

- [ ] Manually tested
- [ ] Unit tested
- [x] Acceptance tested

### Testing

<details open>
  <summary>Testing</summary>
  
</details>

## Further comments


acceptance_tests/database_credential_plan_only_test.go
```diff
@@ -0,0 +1,112 @@
...
+			{
+				Config:             testAccDatabaseCredentialPlanOnlyAdvancedConfig(advancedName),
+				PlanOnly:           true,
+				ExpectNonEmptyPlan: true,
+			},
+			// `access = []` satisfies ExactlyOneOf, so without SizeAtLeast it reaches the
+			// API as a body carrying neither access nor userRoles.
+			{
+				Config:      testAccDatabaseCredentialPlanOnlyEmptyAccessConfig(emptyAccessName),
+				ExpectError: regexp.MustCompile(`(?s)Invalid Attribute Value.*Attribute access list must contain at least 1 elements, got: 0`),
```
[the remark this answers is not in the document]
IsaacLambat: Agreed

internal/datasources/attributes.go
```diff
@@ -96,24 +96,32 @@ func computedStringSet() *schema.SetAttribute {
...
+
 // computedResourcesAttribute builds the resources schema block (buckets.scopes.collections)
-// using Set types. This is reused by both access and privilege schemas.
+// using List types, matching the corresponding resource schemas. This is reused by both access
+// and privilege schemas.
 func computedResourcesAttribute(builder *capellaschema.SchemaBuilder) *schema.SingleNestedAttribute {
 	scopeAttrs := make(map[string]schema.Attribute)
 	capellaschema.AddAttr(scopeAttrs, "name", builder, computedString())
-	capellaschema.AddAttr(scopeAttrs, "collections", builder, computedStringSet())
+	capellaschema.AddAttr(scopeAttrs, "collections", builder, computedStringList())
```
IsaacLambat: Wondering if this is the right solution for this change?

Using List means that order is important, which doesn't seem true in any of these cases. In addition, it also implicitly allows duplicate data which would have to be caught by validators/the API. Set avoids these 2 issues by accepting that the data is unordered, and makes sure there are no duplicates.

internal/resources/database_role_schema.go
```diff
@@ -55,11 +57,16 @@ func DatabaseRoleSchema() schema.Schema {
 		Attributes: resourcesAttrs,
 	})
 
-	capellaschema.AddAttr(attrs, "access", databaseRoleBuilder, &schema.SetNestedAttribute{
+	// SizeAtLeast rejects `access = []`, which satisfies Required but sends a role
+	// with no privileges at all.
+	capellaschema.AddAttr(attrs, "access", databaseRoleBuilder, &schema.ListNestedAttribute{
```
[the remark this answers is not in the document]
IsaacLambat: Was going to put this

Does the API guarantee ordering? If not, drift will happen if the API returns them in a different order since list is an ordered type (unlike Set which is unordered).

docs/resources/database_role.md
```diff
@@ -0,0 +1,90 @@
...
+### Nested Schema for `audit`
+
+Read-Only:
+
+- `created_at` (String) - The RFC3339 timestamp associated with when the resource was initially created.
+ - **Format**: Date-time in RFC3339 format
+- `created_by` (String) - The user who created the resource; this will be a UUID4 ID for standard users and will be a string such as "internal-support" for internal Couchbase support users.
+- `modified_at` (String) - The RFC3339 timestamp associated with when the resource was last modified.
+ - **Format**: Date-time in RFC3339 format
+- `modified_by` (String) - The user who last modified the resource; this will be a UUID4 ID for standard users and wilmal be a string such as "internal-support" for internal Couchbase support users.
```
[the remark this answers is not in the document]
IsaacLambat: Needs to be fixed in the Capella codebase really, not here

internal/resources/database_role_schema.go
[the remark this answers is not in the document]
stanleefdz: One deviation from the suggestion: sorting is right for a Set but wrong for a List — it creates a permanent diff whenever config order differs from sort order (access = [dataWrite, dataRead] would forever plan against sorted [dataRead, dataWrite]). For an order-sensitive List, the fix is aligning the API response to prior state order.

reconcileAccess already had the privilege-set matching map, it just wasn't using it to order output. It now drives from stateAccess, and recurses into new reconcileBuckets/reconcileScopes/reconcileCollections (buckets/scopes matched by name, collections by value). Unmatched API entries are appended, not dropped, so real out-of-band additions still diff. Import path and wildcard suppression unchanged

internal/resources/access_test.go
```diff
@@ -0,0 +1,345 @@
...
+					}
+					out += c.ValueString()
+				}
+				out += " "
+			}
+			out += ") "
+		}
+		out += "]}"
+	}
+	return out + "\n]"
```
IsaacLambat: I know this is a test, but this is not readable nor maintainable

IsaacLambat (commented):
Would it be possible to get some manual testing on this?

stanleefdz:
> Would it be possible to get some manual testing on this?

The acceptance test was tested in local, and logs attached in the Jira

internal/resources/access_test.go
stanleefdz: fixed it in [c40f098](https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/752/commits/c40f098fcb233e84b84ae506ac2c0d66911e32cc)

internal/resources/access.go
```diff
@@ -95,43 +95,109 @@ func mapAccessFromAPI(apiAccess []api.Access) []providerschema.Access {
...
+		func(apiCollection, _ types.String) types.String { return apiCollection })
+}
+
+// reorderToState returns the API elements in prior state order, pairing them by
+// key and combining each matched pair with merge. Elements the API added are
+// appended so the addition still surfaces as a diff; state elements absent from
+// the API are dropped because they no longer exist remotely. A nil slice on
+// either side is returned untouched, preserving Terraform's null/empty
+// distinction. The API elements are never modified in place.
+func reorderToState[T any](apiItems, stateItems []T, key func(T) string, merge func(apiItem, stateItem T) T) []T {
```
IsaacLambat, line 167: Approving on the basis that this has been retested and works

IsaacLambat (approved):

internal/resources/access.go
stanleefdz: yes retested in local and added proofs in jira

PR #753, opened 2026-08-18
[AV-132323] Clean up tainted cluster

l0n3star:
## Jira

* AV-132323

## Description

when a cluster deploy fails, resource is saved to state.  this is what we want, because the resource is marked as tainted and terraform replaces the resource.  we want a replace so that terraform cleans up any infra deployed for the failed cluster.

the problem is the replace never happens.  this PR fixes that

## Type of Change

- [X] Manually tested
- [ ] Unit tested
- [ ] Acceptance tested

### Testing

<details open>
  <summary>Testing</summary>

  validation in ticket

</details>

## Further comments

### DIFF
--- a/internal/api/cluster/state.go
+++ b/internal/api/cluster/state.go
@@ -33,15 +33,12 @@ const (
 type State string
 
 func (s1 State) Equal(s2 State) bool {
-	return strings.EqualFold(string(s1), string(s2))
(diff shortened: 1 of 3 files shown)


l0n3star:
on second thought provider should not try to cleanup infra.  probably better done manually.

PR #755, opened 2026-08-19
[AV-132323] Set state for healthy clusters

l0n3star:
## Jira

* AV-132323

## Description

when cluster deployment fails, the resource is saved to state.  failed cluster resource should not be saved to state.

## Type of Change

- [X] Manually tested
- [ ] Unit tested
- [ ] Acceptance tested

### Testing

<details open>
  <summary>Testing</summary>
  
validation in ticket

</details>

## Further comments

### DIFF
--- a/internal/resources/cluster.go
+++ b/internal/resources/cluster.go
@@ -635,7 +635,8 @@ func (c *Cluster) checkClusterStatus(ctx context.Context, organizationId, projec
 				const msg = "waiting for cluster to complete the execution"
 				tflog.Info(ctx, msg)
 			default:
-				return err
+				tflog.Warn(ctx, "could not get cluster status.  retrying...")
+				continue
 			}
 		}
 	}
--- a/internal/resources/cluster.go
+++ b/internal/resources/cluster.go
@@ -28,14 +28,6 @@ var (
 	_ resource.ResourceWithImportState = &Cluster{}
 )
 
-const errorMessageAfterClusterCreationInitiation = "Cluster creation is initiated, but encountered an error while checking the current" +
-	" state of the cluster. Please run `terraform plan` after 4-5 minutes to know the" +
-	" current status of the cluster. Additionally, run `terraform apply --refresh-only` to update" +
-	" the state from remote, unexpected error: "
-
-const errorMessageWhileClusterCreation = "There is an error during cluster creation. Please check in Capella to see if any hanging resources" +
-	" have been created, unexpected error: "
-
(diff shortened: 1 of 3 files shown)


internal/resources/cluster.go
```diff
@@ -189,42 +181,34 @@ func (c *Cluster) Create(ctx context.Context, req resource.CreateRequest, resp *
...
-	resp.Diagnostics.Append(diags...)
-	if resp.Diagnostics.HasError() {
-		return
-	}
-
-	err = c.checkClusterStatus(ctx, organizationId, projectId, clusterResponse.Id.String())
-	if err != nil {
-		resp.Diagnostics.AddWarning(
+	if err = c.checkClusterStatus(ctx, organizationId, projectId, clusterResponse.Id.String()); err != nil {
+		resp.Diagnostics.AddError(
```
IsaacLambat, line 200: If the cluster is in a failed deployment state, then it will not be added to the state. Is this intentional? The user will have to manually go outside of TF to remove the cluster

internal/resources/cluster.go
```diff
@@ -650,27 +611,32 @@ func (c *Cluster) checkClusterStatus(ctx context.Context, organizationId, projec
...
 			clusterResp, err = c.getCluster(ctx, organizationId, projectId, ClusterId)
 			switch err {
 			case nil:
-				if clusterapi.IsFinalState(clusterResp.CurrentState) {
+				if clusterResp.CurrentState.Equal(clusterapi.Healthy) {
 					return nil
 				}
+
+				if clusterapi.IsTerminalState(clusterResp.CurrentState) {
+					return fmt.Errorf(
```
IsaacLambat, line 630: Should a typed error be returned so callers can distinguish between a failed deployment and an error getting the cluster?

internal/resources/cluster.go
l0n3star: yes that's intentional

as discussed on slack, the point is user may want to contact support to determine why deployment failed, and possibly to remove it in case it is not recoverable.

internal/resources/cluster.go
l0n3star: looking at this closer, we should retry if there is some transient error geting cluster fails 

addressed in https://github.com/couchbasecloud/terraform-provider-couchbase-capella/pull/755/commits/86726d32c4cae3e83b099c553b5b690d13b068e8

stanleefdz (approved):

PR #756, opened 2026-08-21
[AV-139985] Preserve credential_type and user_roles on refresh

stanleefdz:
## Jira

* AV-139985

## Description

The V4 GET response does not include credentialType, so Read derived it from the presence of userRoles: advanced when roles came back, basic otherwise. Any refresh whose response omitted userRoles therefore rewrote an advanced credential's state as basic with null roles, and the RequiresReplace on credential_type turned the next plan into a destroy and recreate that rotated the password - on a routine plan with no configuration change.

Read now carries both fields forward from prior state, as it already does for password and access (SURF-7366). The derivation in retrieveDatabaseCredential is kept for import, where there is no prior state to carry forward; a null
credential_type is what identifies that case.

## Type of Change

- [ ] Manually tested
- [x] Unit tested
- [ ] Acceptance tested

### Testing

<details open>
  <summary>Testing</summary>
  
</details>

## Further comments

### DIFF
--- a/internal/resources/database_credential.go
+++ b/internal/resources/database_credential.go
@@ -270,6 +270,17 @@ func (r *DatabaseCredential) Read(ctx context.Context, req resource.ReadRequest,
 	// For now, we are appending same permissions that the customer passed in the terraform files and not relying on the GET API response.
 	refreshedState.Access = mapAccess(state)
 
+	// credential_type is absent from the GET API response and user_roles cannot be trusted
+	// to be complete, so both are carried forward from prior state rather than re-derived on
+	// every refresh. Deriving credential_type from the response flips an advanced credential
+	// to basic whenever userRoles is missing, and the RequiresReplace on credential_type turns
+	// that into a destroy and recreate (AV-139985). A null credential_type means there is no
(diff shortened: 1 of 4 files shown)


internal/resources/database_credential.go
```diff
@@ -270,6 +270,17 @@ func (r *DatabaseCredential) Read(ctx context.Context, req resource.ReadRequest,
 	// For now, we are appending same permissions that the customer passed in the terraform files and not relying on the GET API response.
 	refreshedState.Access = mapAccess(state)
 
+	// credential_type is absent from the GET API response and user_roles cannot be trusted
+	// to be complete, so both are carried forward from prior state rather than re-derived on
+	// every refresh. Deriving credential_type from the response flips an advanced credential
+	// to basic whenever userRoles is missing, and the RequiresReplace on credential_type turns
+	// that into a destroy and recreate (AV-139985). A null credential_type means there is no
+	// prior state - the import path - where the derived value is the only one available.
```
IsaacLambat: trim this down

IsaacLambat: fix lgtm though

IsaacLambat (approved):

PR #764, opened 2026-08-27
fix: preserve database credential passwords

MishchenkoDaniil:
omit unchanged passwords from database credential update requests
preserve prior password state when no password change is planned
add update request tests for unchanged, changed, and unknown passwords


MishchenkoDaniil:
Validated against the dev environment with a database credential whose Terraform state password intentionally differed from the valid password in Secrets Manager 

- access-only targeted plan: 0 added, 1 changed, 0 destroyed 
- valid current and next credentials remained HTTP 200 after apply 
- stale Terraform-state password remained HTTP 401 
- Secrets Manager AWSCURRENT version was unchanged 
- configuration was restored and the follow-up targeted plan reported no changes 
This confirms the update request omits the stale password while preserving access reconciliation

MishchenkoDaniil:
@panigrahisubhrajit  could u please help with this?

PR #181, opened 2024-03-29
Adding Cluster and App service on/off 

PaulomeeCb:
## Jira

* AV-75849

## Description

This has terraform changes for on-demand cluster and app service on/off. 
TODO: Unit tests and readme.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue). Please, add the "bug" label to the PR.
- [x] New feature (non-breaking change which adds functionality). Please, add the "enhancement" label to the PR.
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected). Please, add the "breaking change" label to the PR.
- [ ] This change requires a documentation update
- [ ] Documentation fix/enhancement

### How was this change tested and do you have evidence? _**(REQUIRED: Select at least 1)**_

- [x] Manually tested
- [ ] Unit tested
- [ ] Acceptance tested
- [ ] Unable to test / will not test (Please provide comments in section below)

### Testing
1. 
**Cluster on-**

Screen Shot 2024-03-29 at 11 33 14 AM
Screen Shot 2024-03-29 at 3 03 37 PM
Screen Shot 2024-03-29 at 11 33 14 AM

linked app service still off - 
Screen Shot 2024-03-29 at 3 06 47 PM

**Cluster on with linked app service turned on-**

Screen Shot 2024-03-29 at 1 08 47 PM
Screen Shot 2024-03-29 at 1 09 07 PM

Screen Shot 2024-03-29 at 1 06 49 PM

**Cluster off-**

NOTE: Make sure to remove the `turnOnLinkedAppservice = true` from request payload while doing this operation.

Screen Shot 2024-03-29 at 4 32 51 PM

Screen Shot 2024-03-29 at 12 34 13 PM

Linked App service automatically turned off-
Screen Shot 2024-03-29 at 12 53 24 PM

2.
**App Service On**

Screen Shot 2024-03-29 at 5 42 20 PM

Screen Shot 2024-03-29 at 5 42 36 PM

**App Service Off**

Screen Shot 2024-03-29 at 5 39 33 PM

Screen Shot 2024-03-29 at 5 39 40 PM

<details open>
  <summary>Testing</summary>
  
</details>

## Required Checklist:

- [ ] I have checked that this change does not generate any credentials and that **they are NOT accidentally logged anywhere**.
- [ ] I have added tests that prove my fix is effective or that my feature works per HashiCorp requirements
- [ ] I have added any necessary documentation (if required)
- [x] I have run make fmt and formatted my code

## Further comments

PR #769, opened 2026-08-31
[AV-142288] Make app service on/off idempotent when already in requested state

nimiyajoseph:
## Jira

* AV-142288

## Description

Applying `couchbase-capella_app_service_onoff_ondemand` with the activation state the app service is *already* in failed the apply. Capella rejects the redundant transition with a `409`, and `manageAppServiceActivation` surfaced it as a hard error — so the resource was not idempotent, and re-applying an unchanged config was never safe.

This absorbs the two "already in the requested state" codes (`11018` for `on`, `11019` for `off`), the same way `manageClusterActivation` already handles `7010`/`7011` for the cluster resource.

One deliberate difference from the cluster resource: Capella returns `11018` for *every* state that forbids a turn-on, including `degraded`, `turnOnFailed` and `deploying`. A blanket swallow of the code would report a broken app service as a successful apply, so the current state is read back and the error is discarded only when it genuinely satisfies the request — `healthy`/`turningOn` for `on`, `turnedOff`/`turningOff` for `off`. A broken app service still fails the apply.

`retrieveAppServiceOnOff` now shares that state read, which also drops a dead `CurrentState` assignment and its `validateAppserviceStateIsSameInPlanAndState` helper.

**Not a regression.** The activation error handling is unchanged in substance since the resource landed in #181 (later edits were cosmetic: `Client` → `ClientV1` in #404, `fmt.Errorf` → `errors.New` in #423), and the server-side guard predates the resource. What is new is `TestAccAppServiceOnOffOnDemandResource`, added in #740 — the first test to drive `Create` against the shared already-healthy app service, which is what exposed a latent bug. Acceptance tests run only on push-to-`main` and manual dispatch, so #740 merged without a green acceptance run.

## Type of Change

- [ ] Manually tested
- [x] Unit tested
- [ ] Acceptance tested

### Testing

<details open>
  <summary>Testing</summary>
  

New table test `Test_appServiceStateSatisfies` covers both directions, the transitional states, and the failure states that must **not** be absorbed (`degraded`, `turnOnFailed`, `turnOffFailed`):

```
$ go test ./internal/resources/ -run Test_appServiceStateSatisfies -v
=== RUN   Test_appServiceStateSatisfies
--- PASS: Test_appServiceStateSatisfies (0.00s)
    --- PASS: Test_appServiceStateSatisfies/on_is_satisfied_by_healthy (0.00s)
    --- PASS: Test_appServiceStateSatisfies/on_is_satisfied_by_turningOn (0.00s)
    --- PASS: Test_appServiceStateSatisfies/on_is_not_satisfied_by_turnedOff (0.00s)
    --- PASS: Test_appServiceStateSatisfies/on_is_not_satisfied_by_degraded (0.00s)
    --- PASS: Test_appServiceStateSatisfies/on_is_not_satisfied_by_turnOnFailed (0.00s)
    --- PASS: Test_appServiceStateSatisfies/off_is_satisfied_by_turnedOff (0.00s)
    --- PASS: Test_appServiceStateSatisfies/off_is_satisfied_by_turningOff (0.00s)
    --- PASS: Test_appServiceStateSatisfies/off_is_not_satisfied_by_healthy (0.00s)
    --- PASS: Test_appServiceStateSatisfies/off_is_not_satisfied_by_turnOffFailed (0.00s)
    --- PASS: Test_appServiceStateSatisfies/unknown_requested_state_is_never_satisfied (0.00s)
    --- PASS: Test_appServiceStateSatisfies/empty_current_state_is_never_satisfied (0.00s)
PASS
ok  	github.com/couchbasecloud/terraform-provider-couchbase-capella/internal/resources	3.764s
```

Full package run, plus the repo's standard checks:

```
$ go test ./internal/resources/
ok  	github.com/couchbasecloud/terraform-provider-couchbase-capella/internal/resources	6.848s

$ goimports -w -local github.com/couchbasecloud/terraform-provider-couchbase-capella <changed files>   # clean
$ gofmt -l <changed files>                                                                             # clean
$ go vet ./internal/resources/ ./acceptance_tests/                                                      # clean
$ go build -ldflags "-s -w -X '...ProviderVersion=v1.10.0'" -o ./bin/terraform-provider-couchbase-capella   # ok
```

</details>

## Further comments

`TestAccAppServiceOnOffOnDemandAlreadyInRequestedState` is added in this PR but has **not** been executed — it needs a live Capella org with a healthy app service, and the acceptance suite only runs on push-to-`main` or manual dispatch. It applies `state = "on"` against the already-healthy global app service (the exact case that was failing), then re-plans to assert the plan is empty. The pre-existing `TestAccAppServiceOnOffOnDemandResource` exercises the same path and should go green on the next acceptance run.

Out of scope, flagged for a separate ticket: `Read` echoes the stored `state` back rather than the app service's real state, so an app service turned off outside Terraform still reads as `on`. This PR does not change that behaviour.

### DIFF
--- a/internal/resources/appservice_onoff.go
+++ b/internal/resources/appservice_onoff.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
-	"strings"
 
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
@@ -33,6 +32,13 @@ const errorMessageAfterAppServiceOnOffCreation = "Cluster switch on/off is succe
 	" state of the switched on/off cluster. Please run `terraform plan` after 1-2 minutes to know the" +
 	" current state. Additionally, run `terraform apply --refresh-only` to update" +
 	" the state from remote, unexpected error: "
+
+// Capella rejects an activation state change with a 409 when the app service is not in a
+// state that permits it, which includes it already being in the requested state.
+const (
+	errCodeAppServiceNotOffToBeTurnedOn = 11018
+	errCodeAppServiceNotOnToBeTurnedOff = 11019
+)
 
 // AppServiceOnOffOnDemand is the AppServiceOnOffOnDemand implementation.
 type AppServiceOnOffOnDemand struct {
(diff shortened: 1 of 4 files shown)


internal/resources/appservice_onoff.go
```diff
@@ -160,11 +166,81 @@ func (a *AppServiceOnOffOnDemand) manageAppServiceActivation(ctx context.Context
...
+
+// appServiceStateSatisfies reports whether currentState already satisfies the requested on/off
+// state. In-progress transitions count as satisfied so that a state read racing with a
+// concurrent transition does not turn a no-op into an apply failure.
+func appServiceStateSatisfies(state string, currentState apps_service_api.State) bool {
+	switch state {
+	case "on":
+		return currentState == apps_service_api.Healthy || currentState == apps_service_api.TurningOn
+	case "off":
+		return currentState == apps_service_api.TurnedOff || currentState == apps_service_api.TurningOff
```
IsaacLambat, line 214: Do we need `TurnOnFailed` and `TurnOffFailed` here? When these states are entered, we shouldn't really be automatically retrying

nimiyajoseph: Those lines are the transitional states rather than the failed ones — `TurningOn` / `TurningOff`:

```go
case "on":
	return currentState == apps_service_api.Healthy || currentState == apps_service_api.TurningOn
case "off":
	return currentState == apps_service_api.TurnedOff || currentState == apps_service_api.TurningOff
```

`TurnOnFailed` / `TurnOffFailed` don't appear in this file at all — the only place they occur is the unit test, as cases asserting `false`. The names are a letter apart in the middle, so they read very similarly.

And we agree on the behaviour. For an app service in `turnOnFailed` with `state = "on"`:

1. `POST .../activationState` → server-side `TurnOn` falls to its `default:` branch → `409` / code `11018`
2. `appServiceAlreadyInState` matches the code, so it reads the current state back → `turnOnFailed`
3. `appServiceStateSatisfies("on", turnOnFailed)` → `false`
4. the original `409` is returned unchanged, and the apply fails with "App Service activation failed"

No retry, no absorption. Avoiding exactly that case — reporting success over a broken app service — is why this re-reads the state instead of blanket-swallowing the code the way `manageClusterActivation` does for `7010`/`7011`. It's pinned by `on is not satisfied by turnOnFailed` and `off is not satisfied by turnOffFailed` in `appservice_onoff_test.go`.

On the states that *are* listed: `TurningOn`/`TurningOff` cover a narrow race where the `409` is raised at POST time and, by the time we read the state back, a concurrent actor has started the transition we wanted. Counting those as satisfied is consistent with the resource's existing contract, since the happy path doesn't wait for completion either — the API returns `202 Accepted` and the resource returns immediately, so a successful apply never implied `healthy`.

Happy to tighten it to `Healthy`/`TurnedOff` only if you'd prefer — that trades the race for an occasional spurious apply failure.

IsaacLambat (approved):

internal/resources/appservice_onoff.go
IsaacLambat: Ahh, I meant that every time the user runs `terraform apply` and the App Service is in a failed state, this activation state request will be retried. But then I don't want to hide the fact it failed so LGTM.

PR #798, opened 2026-09-17
[AV-143188] Add acceptance tests for bucket backup export resource

karanjain-ops:
## Jira

* AV-143188

Stacked on #797, which adds the resource these tests exercise. Review that first.

## Description

Adds `acceptance_tests/bucket_backup_export_acceptance_test.go`, the acceptance test suite for the `couchbase-capella_bucket_backup_export` resource.

- `TestAccBucketBackupExportResource` declares a bucket, a backup of it and an export of that backup in one config. Terraform orders them by their references and the `couchbase-capella_backup` resource already blocks until the backup is `ready`, so the test needs no polling of its own. Covers create, read and import, which is the resource's whole lifecycle.
- Runs on a bucket created per test run rather than the shared `globalBucketId`. Capella serialises manual backups per bucket and allows one active export per cycle, so a leaked export from a killed run would fail the test with an unrelated 409.
- Identity is asserted with `TestCheckResourceAttrPair` against the bucket and backup resources rather than against literals.
- Error cases match on the goof code rather than the message: Terraform hard-wraps diagnostics mid-phrase, so message substrings do not match reliably, while the codes are stable and part of the public contract.
- The duplicate-cycle test uses `depends_on` to order the two exports; without it they race and the test is flaky.
- No update step, since the resource does not support update.
- The archive attributes are not asserted. The export is still `pending` when the apply returns, so all four are null at that point, and reaching `complete` would tie the suite's runtime to the backup infrastructure. That path is covered by the evidence on #797.

| Test | What it guards |
|---|---|
| `TestAccBucketBackupExportResource` | create, read, import |
| `TestAccBucketBackupExportResourceDuplicateCycle` | second export for a cycle returns 409 |
| `TestAccBucketBackupExportResourceUnknownBackup` | unknown backup returns 404/5017 |
| `TestAccBucketBackupExportResourceUnknownCluster` | unknown cluster returns 404/4025 |
| `TestAccBucketBackupExportResourceWrongProject` | project not owning the cluster returns 422/4031 |
| `TestAccBucketBackupExportResourceMalformedBackupId` | non-UUID rejected at plan time |
| `TestAccBucketBackupExportResourceEmptyBucketId` | empty `bucket_id` rejected at plan time |

## Type of Change

- [ ] Manually tested
- [ ] Unit tested
- [x] Acceptance tested

### Testing

<details open>
  <summary>Testing</summary>

The P0 Sanity Suite runs the acceptance tests in CI, but only the ones named in `acceptance_tests/sanity.list`. The second commit adds `TestAccBucketBackupExportResource` to that list, so it now runs there. I cannot run the suite locally, as I do not have `TF_VAR_host` and `TF_VAR_auth_token` for the CI tenant.

Locally it compiles, vets and formats clean:

```
$ go vet ./acceptance_tests/
$ go test -c -o /dev/null ./acceptance_tests
$ gofmt -l acceptance_tests/bucket_backup_export_acceptance_test.go
$ goimports -l -local github.com/couchbasecloud/terraform-provider-couchbase-capella acceptance_tests/bucket_backup_export_acceptance_test.go
```

All four produce no output and exit zero.

</details>

## Further comments

Please run the suite against the CI tenant before merging. The two things most likely to behave differently there are the per-bucket backup serialisation, which is why the test creates its own bucket, and the duplicate-cycle test, which needs the first export still active when the second is attempted.


stanleefdz (dismissed):

saiakhil2012 (approved):
Re-approving after a base change in PR Stack

